PR #24, opened 2022-10-24
WIP: Add Binding feature

lukx:
Work in progress:

Some commands require the two sides in the relation to bind, this is an attempt to allow our end to start the binding.

### DIFF
--- a/spine/binding_manager.go
+++ b/spine/binding_manager.go
@@ -1,0 +1,149 @@
+package spine
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"sync/atomic"
+
+	"github.com/DerAndereAndi/eebus-go/spine/model"
+	"github.com/ahmetb/go-linq/v3"
+)
+
+type BindingManager interface {
+	AddBinding(localDevice *DeviceLocalImpl, remoteDevice *DeviceRemoteImpl, data model.BindingManagementRequestCallType) error
+	RemoveBinding(data model.BindingManagementDeleteCallType, remoteDevice *DeviceRemoteImpl) error
+	Bindings(remoteDevice *DeviceRemoteImpl) []*BindingEntry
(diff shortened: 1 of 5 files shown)


DerAndereAndi:
Do you have an example to such a command that requires both sides to bind? Maybe a reference to the spec? I am curious :)

I only came across a scenario yet in which a write command requires a binding, but that is one sided only.

lukx:
The Binding is known to both parties, but is initiated by one side. So my wording may have been imprecise. 
So, from my understanding of the Binding Spec in `EEBus_SPINE_TS_ProtocolSpecification.pdf Z 2124ff` the writing party can start a binding, but the receiving party has to accept it.
In my example:
The CEM implementation will add a binding to the Hvac Feature of the Heatpump. For this Feature, the Heatpump carries the Role "Server". 
The Heatpump registers the binding; so does the CEM.
Both parties seem to be able to revoke the binding.(L2161), so I derive that both parties need to keep track of the binding.

DerAndereAndi:
If the receiver declines the binding, then writing is not allowed and will not work. The CEM is the one that needs to register a binding with the HeatPump (in your case), if the heatpump declines, the cem may not sent write messages. The heatpump does not register a binding to the server.

The CEM may only send write messages, if the corresponding binding was accepted by the heatpump.
The Heatpump may only accept write messages, if the CEM registered a binding and the heatpump accepted it.

lukx:
Correct, and that's also what I observe from my heatpump (well, the non-accepting part :-/ - have not yet successfully bound)

DerAndereAndi:
The method to send the binding is not yet part of this PR. Something similar to `SubscribeAnWait(...` in `FeatureLocal`

DerAndereAndi:
Ich merge den PR mal as is, es ist ein Anfang und man kann darauf sehr gut aufbauen. Muss ja nicht immer alles in einem PR "final" sein ;) Danke @lukx !

PR #48, opened 2023-01-26
Experiment with heatpumps

DerAndereAndi:
This is for experimentation only!

### DIFF
--- a/cmd/hems/main.go
+++ b/cmd/hems/main.go
@@ -127,57 +127,65 @@ func (h *hems) HandleEvent(payload spine.EventPayload) {
 func (h *hems) heatpumpConnected(entity *spine.EntityRemoteImpl) {
 	localDevice := h.myService.LocalDevice()
 
-	deviceClassification, _ := features.NewDeviceClassification(model.RoleTypeClient, model.RoleTypeServer, localDevice, entity)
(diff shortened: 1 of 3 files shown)


wimleers:
Very, very interested in this. I'm an owner of a Vaillant 75/5 since last week. Happy to work with you on this! (And got things started over at https://github.com/enbility/eebus-go/issues/54#issuecomment-1522871572.)

sylvainvh:
Hello @DerAndereAndi , I checked your PR but this branch seems really outdated. I tried to rebase your branch with dev but I've to much conflict.
I'm a developper (python) but I never code in Go (Go seems to be easy to understand). However, I can help you with some tests if you want.

DerAndereAndi:
Hi @sylvainvh, 

yeah I didn't continue as these old devices don't reveal the data needed to be able to communicate further with them via EEBUS. They use an very old specficiation version, but even with that one it seems internal know how is needed to use them.

In general, the implemented use cases don't offer a lot of possibilities other than shifting power usage a bit (roughly speaking). So the possibilities to control these devices are very very limited. Anyhow, if I somehow find out more about the internal, I surely would continue this PR to get at least that working.

Best,
Andreas

PR #88, opened 2024-06-23
Add workaround for invalid MEB measurements

DerAndereAndi:
The PMCC EVSE can connect to MEB EVs via ISO and then uses the EV as NTP server. Sadly that server reports the wrong time and thus all measurements are invalid.

This commit adds a workaround to test if the measurement is within 1 minute if the time is shifted by 1 or 2 hours.


andig:
Why should a network-connected PMCC even use the vehicle as NTP source? That seems like a very stupid idea in the first place...

PR #97, opened 2024-07-28
Update service.go

andig:
Prevent double space


DerAndereAndi:
I added this change in another cleanup commit. Thanks for the hint

PR #98, opened 2024-07-28
Simplify service start

andig:


DerAndereAndi:
I'd like to keep it as is. Thanks for the idea.

PR #105, opened 2024-08-15
Fix potential race condition leading to nil dereference

### LINKED ISSUE #104: Potential RaceCondition in CS-LPC and CS-LPP in WriteApprovalCallbacks
When a Feature for a LocalEntity has multiple WriteApprovalCallbacks (e.g. LPC+LPP) and a command that requires write approval is sent for one of these features, depending on the order in which the WriteApprovalCallbacks are executed a nil pointer dereference can occur causing a segmentation fault.

In general the segmentation fault occurs when the ApproveOrDenyWrite method is called on a ship-go FeatureLocalInterface, the passed *Message is not complete (i.e. contains only a valid MsgCounter) and "completes" the number of required approvals such that processWrite is then called with an incomplete *Message.

One potential fix for this issue would be splitting the public ApproveOrDenyProductionLimit messages into a public and private part where the private part directly takes a *shipapi.Message which it can pass to ApproveOrDenyWrite and can be called from the WriteApprovalCallback. Another potential fix would be to adapt ApproveOrDenyWrite in ship-go to cache the "most valid" message and pass that through to processWrite instead of passing the last message.

If we decide on the first option, we should definitely document the API contract for ApproveOrDenyWrite that it only accepts "valid" messages and may crash when passed incomplete Messages.

### Backtrace

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x8ca891]

goroutine 181 [running]:
github.com/enbility/spine-go/spine.(*FeatureLocal).processWrite(0xc000000300, 0xc000632008)
        /home/simon/projects/intern/02_eebus/eebus-servers/vendor/github.com/enbility/spine-go/spine/feature_local.go:674 +0xb1
github.com/enbility/spine-go/spine.(*FeatureLocal).ApproveOrDenyWrite(0xc000000300, 0xc000632008, {0x0, 0x0})
        /home/simon/projects/intern/02_eebus/eebus-servers/vendor/github.com/enbility/spine-go/spine/feature_local.go:236 +0x3e5
github.com/enbility/eebus-go/usecases/cs/lpp.(*LPP).ApproveOrDenyProductionLimit(0xc0002b2cc0, 0x128, 0x1, {0x0, 0x0})
        /home/simon/projects/intern/02_eebus/eebus-servers/vendor/github.com/enbility/eebus-go/usecases/cs/lpp/public.go:151 +0x246
created by github.com/enbility/eebus-go/usecases/cs/lpp.(*LPP).loadControlWriteCB in goroutine 122
        /home/simon/projects/intern/02_eebus/eebus-servers/vendor/github.com/enbility/eebus-go/usecases/cs/lpp/usecase.go:148 +0x325
```

### Illustrative Example

1. A device registers the usecases cs/lpc and cs/lpp which both register their loadControlWriteCBs.
2. The device receives an LPC command and the WriteApprovalCallbacks for both use cases are entered in separate goroutines "simultaneously" ([LPC](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpc/usecase.go#L110), [LPP](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpp/usecase.go#L109))
3. The LPC handler proceeds through [line 140](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpc/usecase.go#L141) and calls the EventCB registered on the LPC use-case.
4. The LPC EventCB calls [ApproveOrDenyConsumptionLimit](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpc/public.go#L133) which passes through to line 159 and calls [ApproveOrDenyWrite](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L196)
5. ApproveOrDenyWrite runs through until [line 215](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L215) where since it's the first handler out of two handlers total it sets `r.writeApprovalReceived[*msg.RequestHeader.MsgCounter] = 1` and returns (1 < 2 [line 222](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L222))
6. The LPP handler then proceeds through to [line 148](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpp/usecase.go#L148) because limitId != item.LimitId for all entries in the LoadControlLimitData and calls [ApproveOrDenyProductionLimit](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpp/public.go#L133) with the MsgCounter (and only the MsgCounter of the message)
7. ApproveOrDenyProductionLimit continues until [line 143](https://github.com/enbility/eebus-go/blob/b779bca324be309acf8b053ef7432dbad56c0418/usecases/cs/lpp/public.go#L143) where it enters the !ok branch because there are no pending LPP limits.
8. ApproveOrDenyProductionLimit then creates a new spineapi.Message which only has RequestHeader.MsgCounter set (all other fields are nil) and calls [ApproveOrDenyWrite](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L196)
9. ApproveOrDenyWrite continues until [line 215](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L215) sets `r.writeApprovalReceived[*msg.RequestHeader.MsgCounter] = 1 + 1` passes through the check on [line 222](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L222) (2 >= 2) and continues until [line 236](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L236) to call [processWrite](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L672)
10. processWrite calls [executeWrite](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L683) which calls `msg.Cmd.Data()` which returns `errors.New("Data not found in Cmd")` because Cmd is zero-initialized and all its fields are nil.
11. executeWrite then returns `model.NewErrorType(model.ErrorNumberTypeCommandNotSupported, err.Error())` and execution continues in processWrite on [line 674](https://github.com/enbility/spine-go/blob/4490d86aee74200e897ad17e105e9e46c5f8b4b2/spine/feature_local.go#L674) with a call to `msg.FeatureRemote.Device().Sender().ResultError(msg.RequestHeader, r.Address(), err)` which proceeds to segfault on calling the Device() method of the FeatureRemote interface which was zero-initialized to nil.

This segfault also occurs when the LPP handler completes first for an LPP command (then segfaulting during the LPC-handler). The segfault does not occur when the LPC handler completes second for an LPC command or the LPP handler completes second for an LPP command as then the actual message is retrieved from the pendingLimits map and passed to ApproveOrDenyWrite.

sthelen-enqs:
Split ApproveOrDeny*Limit public functions into a public part which takes a msgCounter and resolves it to a Message using the pendingLimits as before and a private part which takes a *Message and can be called directly from the internal WriteApprovalCallbacks. This removes the need to call FeatureLocalInterface.ApproveOrDenyWrite with an "empty" message when a message is not found in the pendingLimits for ApprovalRequests that don't match the current use case.

Fixes #104

### DIFF
--- a/usecases/cs/lpc/public.go
+++ b/usecases/cs/lpc/public.go
@@ -7,7 +7,6 @@ import (
 	"github.com/enbility/eebus-go/api"
 	"github.com/enbility/eebus-go/features/server"
 	ucapi "github.com/enbility/eebus-go/usecases/api"
-	spineapi "github.com/enbility/spine-go/api"
 	"github.com/enbility/spine-go/model"
 	"github.com/enbility/spine-go/util"
 )
@@ -134,29 +133,13 @@ func (e *LPC) ApproveOrDenyConsumptionLimit(msgCounter model.MsgCounterType, app
 	e.pendingMux.Lock()
 	defer e.pendingMux.Unlock()
 
-	f := e.LocalEntity.FeatureOfTypeAndRole(model.FeatureTypeTypeLoadControl, model.RoleTypeServer)
-
-	result := model.ErrorType{
-		ErrorNumber: model.ErrorNumberType(0),
-	}
-
 	msg, ok := e.pendingLimits[msgCounter]
 	if !ok {
-		// it is not a limit of this usecase, so approve it
-		newMsg := &spineapi.Message{
-			RequestHeader: &model.HeaderType{
-				MsgCounter: &msgCounter,
-			},
-		}
-		f.ApproveOrDenyWrite(newMsg, result)
+		// no pending limit for this msgCounter, this is a caller error
 		return
 	}
 
-	if !approve {
-		result.ErrorNumber = model.ErrorNumberType(7)
-		result.Description = util.Ptr(model.DescriptionType(reason))
-	}
-	f.ApproveOrDenyWrite(msg, result)
+	e.approveOrDenyConsumptionLimit(msg, approve, reason)
 
 	delete(e.pendingLimits, msgCounter)
 }
--- a/usecases/cs/lpc/usecase.go
+++ b/usecases/cs/lpc/usecase.go
@@ -103,6 +103,20 @@ func (e *LPC) loadControlServerAndLimitId() (lc *server.LoadControl, limitid mod
 	return lc, *description.LimitId, nil
 }
 
+func (e *LPC) approveOrDenyConsumptionLimit(msg *spineapi.Message, approve bool, reason string) {
+	f := e.LocalEntity.FeatureOfTypeAndRole(model.FeatureTypeTypeLoadControl, model.RoleTypeServer)
+
+	result := model.ErrorType{
+		ErrorNumber: model.ErrorNumberType(0),
+	}
+
+	if !approve {
+		result.ErrorNumber = model.ErrorNumberType(7)
+		result.Description = util.Ptr(model.DescriptionType(reason))
+	}
+	f.ApproveOrDenyWrite(msg, result)
+}
+
 // callback invoked on incoming write messages to this
 // loadcontrol server feature.
 // the implementation only considers write messages for this use case and
@@ -145,7 +159,7 @@ func (e *LPC) loadControlWriteCB(msg *spineapi.Message) {
 	e.pendingMux.Unlock()
 
 	// approve, because this is no request for this usecase
-	go e.ApproveOrDenyConsumptionLimit(*msg.RequestHeader.MsgCounter, true, "")
+	go e.approveOrDenyConsumptionLimit(msg, true, "")
 }
 
 func (e *LPC) AddFeatures() {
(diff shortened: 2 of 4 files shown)


DerAndereAndi:
Thanks a lot for the great bug report and PR!

PR #107, opened 2024-08-29
Fix ma/mpc usecase to validate the correct actor type

sthelen-enqs:
The ma/mpc usecase should filter for actors of type MonitoredUnit but was filtering for actors of type MonitoringAppliance which lead to the uc never finding any entities.

This PR also adds a commit which adjusts the string representation of the ma/mpc UseCaseSupportUpdate EventType from "ma-mpc-oscev-UseCaseSupportUpdate" to "ma-mpc-UseCaseSupportUpdate" as it isn't relevant here. This is technically an API-breaking change, but all subscribers should be using the constant mpc.UseCaseSupportUpdate instead of its string representation and therefore automatically always use the correct variant for their version of eebus-go. I could, however, drop the commit from this PR if preferred.

### DIFF
--- a/usecases/ma/mpc/types.go
+++ b/usecases/ma/mpc/types.go
@@ -6,7 +6,7 @@ const (
 	// Update of the list of remote entities supporting the Use Case
 	//
 	// Use `RemoteEntities` to get the current data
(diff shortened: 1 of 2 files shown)


DerAndereAndi:
Nice catch @sthelen-enqs . Thank you!

PR #108, opened 2024-08-29
Implement the MPC use case for Monitored Units

sthelen-enqs:
This adds an implementation of the MPC use case for Monitored Units.

The current implementation has 2 major restrictions so far:
1. it automatically configures itself as a 3-phase meter (ABC) and there's no way to specify a different configuration right now
2. it only supports the required phase-to-neutral and not the phase-to-phase voltages

I'm opening this as a Draft PR because while this UC works, it can still be improved and I'm looking for feedback and potential improvements. Some of the areas I'm explicitly looking for feedback in are:
- how best to make instantiation of the use case configurable (number of phases supported etc)
- how best to adapt the public usecase API when phases are configurable (e.g. adapting functions to take []float64 or something else entirely)
- how to improve the current tests, I thought about connecting a remoteDevice and reading the set values out via that but couldn't get that to work easily, the alternative as far as I can see would be to add functions to the public api and copy the [cs/lpc tests](https://github.com/enbility/eebus-go/blob/22e8d65b137563879d62c530ef44ffb040f16b2e/usecases/cs/lpc/public_test.go#L155)


DerAndereAndi:
Thanks a lot for the PR!

Some first thoughts, more will surely come:

Having individual public functions for each scenario is what surely is expected and needed. But using those individually will result in a notify update message for each call, which I think is not nice.

I do see 2 options here:
- Add another public method that can take all measurement values for all scenarios, with all being optional at the end so only the ones with actual data will be used
- Introducing a mechanism in the SPINE repository to combine multiple local data updates to work like a transaction, where only the transaction end will trigger the actual local data storage update which will then automatically trigger the notify messages to all subscribers

DerAndereAndi:
Regarding initialization: how about the possibility to define the supported scenarios, and for each scenario define a configuration item where items like amount of phases, measurement type, reference to etc. needs to be set.

Scenario 1 is required, all other scenarios are option/recommended. And even within the scenarios there are optional items, e.g. phase specfiic power.

Also I am wondering you only added setters, but maybe it is also helpful to have getters to e.g. validate the current storage?

usecases/api/mu_mpc.go
```diff
@@ -0,0 +1,108 @@
...
+	//   - phaseB: the active power of phase B
+	//   - phaseC: the active power of phase C
+	SetPowerPerPhase(phaseA, phaseB, phaseC float64) error
+
+	// return the momentary active phase specific power consumption or production per phase
+	//
+	// possible errors:
+	//   - ErrDataNotAvailable if no such values are (yet) available
+	//   - and others
+	PowerPerPhase() ([]float64, error)
```
DAMEK86, line 40: it is obvious that the returned slice contains values for phases A, B, C but I would prefer adding some phase identifiers as well.
With this change, you can provide additional functionality like searching for a specific phase or checking if the value in a particular phase is available if only 1 or 2 values of 3 or more are returned.

DerAndereAndi: The current implementation aligns with the API structure of the other use cases. If this should be changed, it should be changed everywhere. See e.g.  the CEM use case implementations e.g. https://github.com/enbility/eebus-go/blob/dev/usecases/api/cem_opev.go

DAMEK86: Hmm from a user of this lib point of view this could be improved but agreed, this is then not part of this Pull Request.

Thx for clarifying

usecases/mu/mpc/public.go
```diff
@@ -0,0 +1,305 @@
...
+	err = e.setMeasurementDataForId(e.acCurrent[1], phaseB)
+	if err != nil {
+		return err
+	}
+
+	err = e.setMeasurementDataForId(e.acCurrent[2], phaseC)
+	if err != nil {
+		return err
+	}
+
```
DerAndereAndi, line 127: Calling individual setters will trigger individual partial update notifications. This is just one of multiple situations where this happens. I think we need a better solution to handle updates of multiple data points in a single step.

Same if someone wants to update e.g. currents and voltages, should be possible in one step as well.

sthelen-enqs:
I've updated the PR with support for updating multiple measurements at the same time using the Update() function. For an example as to how to use the Update() function look at [the Test_PowerPerPhase test](https://github.com/enbility/eebus-go/pull/108/files#diff-154e28c3e4fe1bcdf6b49144b13313838cea5d951dd0cf8f75a5f7ba15dc2bd7R21).

The mu/mpc use case is now also configurable for e.g. non-three-phase use cases by passing configuration parameters to the NewMpc function. The config types are located in usecases/mu/mpc/config.go. For an example as to how to use these configuration parameters, look at [the MPC testhelper](https://github.com/enbility/eebus-go/pull/108/files#diff-78ee22b82c0a2ca9d83370101d1833ea886be629ba34ad6f502d916a7144fd34R77)

usecases/usecase/usecase.go
```diff
@@ -88,9 +91,14 @@ func (u *UseCaseBase) UpdateUseCaseAvailability(available bool) {
 
 func (u *UseCaseBase) IsCompatibleEntityType(entity spineapi.EntityRemoteInterface) bool {
 	if entity == nil {
+
```
DerAndereAndi: The linter is reporting `unnecessary leading newline (whitespace)`. Would be great to change this.

usecases/mu/mpc/config.go
```diff
@@ -0,0 +1,95 @@
...
+type ConnectedPhases string
+
+const ConnectedPhasesA ConnectedPhases = "a"
+const ConnectedPhasesB ConnectedPhases = "b"
+const ConnectedPhasesC ConnectedPhases = "c"
+const ConnectedPhasesAB ConnectedPhases = "ab"
+const ConnectedPhasesBC ConnectedPhases = "bc"
+const ConnectedPhasesCA ConnectedPhases = "ac"
+const ConnectedPhasesABC ConnectedPhases = "abc"
+
```
DerAndereAndi, line 11: I usually write such a set like this:

```
const (
  ConnectedPhasesA ConnectedPhases = "a"
  ConnectedPhasesB ConnectedPhases = "b"
  ConnectedPhasesC ConnectedPhases = "c"
  ...
)
```

For consistency this may be a good idea here as well? What do you think?

usecases/mu/mpc/config.go
```diff
@@ -0,0 +1,95 @@
...
+// MonitorPowerConfig is the configuration for the monitor use case
+// This config is required by the mpc use case and must be used in mpc.NewMPC
+type MonitorPowerConfig struct {
+	ConnectedPhases ConnectedPhases // The phases that are measured
+
+	ValueSourceTotal  *model.MeasurementValueSourceType // The source of the values from the acPowerTotal (not optional)
```
DerAndereAndi: Instead of `(not optional)`, how about writing `required` or `mandatory`?

usecases/mu/mpc/config.go
```diff
@@ -0,0 +1,95 @@
...
+
+// MonitorCurrentConfig is the configuration for the monitor use case
+// If this config is passed via NewMPC, the use case will support current monitoring
+// The current phases will be the same as specified in MonitorPowerConfig
+type MonitorCurrentConfig struct {
+	ValueSourcePhaseA *model.MeasurementValueSourceType // The source of the values for phase A (shall be set if the phase is supported)
+	ValueSourcePhaseB *model.MeasurementValueSourceType // The source of the values for phase B (shall be set if the phase is supported)
+	ValueSourcePhaseC *model.MeasurementValueSourceType // The source of the values for phase C (shall be set if the phase is supported)
+
+	ValueConstraintsPhaseA *model.MeasurementConstraintsDataType // The constraints for the current for phase A (optional can be nil) (needs ValueSourcePhaseA to be set)
```
DerAndereAndi: Instead of ' needs ... to be set', how about ' requires ... to be set'  ?

usecases/mu/mpc/config.go
```diff
@@ -0,0 +1,95 @@
...
+const ConnectedPhasesCA ConnectedPhases = "ac"
+const ConnectedPhasesABC ConnectedPhases = "abc"
+
+// MonitorPowerConfig is the configuration for the monitor use case
+// This config is required by the mpc use case and must be used in mpc.NewMPC
+type MonitorPowerConfig struct {
+	ConnectedPhases ConnectedPhases // The phases that are measured
+
+	ValueSourceTotal  *model.MeasurementValueSourceType // The source of the values from the acPowerTotal (not optional)
+	ValueSourcePhaseA *model.MeasurementValueSourceType // The source of the values from the acPower for phase A (shall be set if the phase is supported)
```
DerAndereAndi: The SPINE spec uses `SHALL`, I am not sure if the meaning of the wording is always clear. How about using `required to be set, ...` instead?

usecases/mu/mpc/usecase.go
```diff
@@ -0,0 +1,574 @@
...
+	acPowerTotal     *model.MeasurementIdType
+	acPower          [3]*model.MeasurementIdType
+	acEnergyConsumed *model.MeasurementIdType
+	acEnergyProduced *model.MeasurementIdType
+	acCurrent        [3]*model.MeasurementIdType
+	acVoltage        [6]*model.MeasurementIdType // Phase to phase voltages are not supported (yet)
+	acFrequency      *model.MeasurementIdType
+}
+
+func NewMPC(
```
DerAndereAndi, line 49: I haven't done this yet consistently either, but I'd like to improve this overall by adding documentation on the `New...` methods as this has to be used by every application and it should be clear what is meant by which parameter. Especially when adding the configuration parameters here, which is new to all.

usecases/mu/mpc/usecase.go
```diff
@@ -0,0 +1,574 @@
...
+		},
+		{
+			Scenario:  model.UseCaseScenarioSupportType(5),
+			Mandatory: false,
+			ServerFeatures: []model.FeatureTypeType{
+				model.FeatureTypeTypeElectricalConnection,
+				model.FeatureTypeTypeMeasurement,
+			},
+		},
+	}
```
DerAndereAndi, line 116: The scenarios should only be added, if the configuration for this scenario is provided and hence the application supports it. Right now this code would always report that all scenarios are supported.

usecases/mu/mpc/usecase.go
```diff
@@ -0,0 +1,574 @@
...
+			constraints = append(constraints, *e.frequencyConfig.ValueConstraints)
+		}
+	}
+
+	electricalConnection, err := server.NewElectricalConnection(e.LocalEntity)
+	if err != nil {
+		panic(err)
+	}
+
+	idEc1 := model.ElectricalConnectionIdType(0)
```
DerAndereAndi: This will cause problems, as other use cases may use an eletrical connection in the entity, and they may not be identical.

Instead we should check if one exists with the properties defined in the spec and use that. Otherwise create one.

usecases/mu/mpc/testhelper_test.go
```diff
@@ -0,0 +1,118 @@
...
+				ValueRangeMin: model.NewScaledNumberType(0),
+				ValueRangeMax: model.NewScaledNumberType(100),
+				ValueStepSize: model.NewScaledNumberType(1),
+			}),
+		},
+	)
+	s.sut.AddFeatures()
+	s.sut.AddUseCase()
+
+	//s.remoteDevice, s.monitoredEntity = setupDevices(s.service, s.T())
```
DerAndereAndi: I would love to see such tests as in CS/LPC to be added, not just using the mocks.

DerAndereAndi (changes requested):
Awesome updates, I really like the way you solve the multiple datapoint update topic!

usecases/mu/mpc/config.go
andig: fwiw: you can also use enumer to have these created and convert from/to string.

sthelen-enqs: updated and good to know, thanks

usecases/mu/mpc/usecase.go
sthelen-enqs: I've added a basic doc comment to the function, is this good or should we go more in depth?

usecases/mu/mpc/usecase.go
sthelen-enqs: Should be fixed now, though it might be nice to have tests for it as well.

sthelen-enqs:
I've updated the PR with fixes for the simple changes, will look over the rest when I have more time

usecases/mu/mpc/usecase.go
DerAndereAndi: This looks good to me. I also started adding comments to the other implementations in https://github.com/enbility/eebus-go/pull/136

usecases/mu/mpc/usecase.go
DerAndereAndi: Yes. Test coverage still has potential to grow ;)

usecases/mu/mpc/config.go
```diff
@@ -0,0 +1,98 @@
...
+	ValueConstraintsPhaseCToA *model.MeasurementConstraintsDataType // The constraints for the voltage for phase C to A (optional can be nil) (requires ValueSourcePhaseCToA to be set)
+}
+
+// MonitorFrequencyConfig is the configuration for the monitor use case
+type MonitorFrequencyConfig struct {
+	ValueSource      *model.MeasurementValueSourceType     // The source of the values (required)
+	ValueConstraints *model.MeasurementConstraintsDataType // The constraints for the frequency values (optional can be nil)
+}
+
+func (c *MonitorPowerConfig) SupportsPhases(phase []string) bool {
```
DerAndereAndi, line 58: It would be great to have some tests for this method standalone as well. So line 93 is also covered.

usecases/mu/mpc/usecase.go
```diff
@@ -0,0 +1,605 @@
...
+		voltageConfig:   monitorVoltageConfig,
+		frequencyConfig: monitorFrequencyConfig,
+	}
+
+	_ = spine.Events.Subscribe(uc)
+
+	return uc, nil
+}
+
+func (e *MPC) AddFeatures() {
```
DerAndereAndi: This method is getting quite loooooooong. How about splitting it up into individual functions for each config?
Also I am still not sure if using panics is the right approach. Maybe `AddFeatures()` signature should be changed to `AddFeatures() error` overall ?

Also the code looks quite repetitive, and maybe there is a good way to shrink this down, and thus making it also easire to improve the test coverage for all error scenarios?

DerAndereAndi (changes requested):
Great improvements already, thank you!

usecases/mu/mpc/usecase.go
sthelen-enqs: We added a GetOrAddIdForDescription function to the electricalConnection which checks if there's an ID we can reues and otherwise creates a new id

usecases/mu/mpc/testhelper_test.go
sthelen-enqs: We updated the tests to use Filters to query the stored data representation for the usecases. This should test that querying the data works from the client perspective without needing to reimplement the ma/mpc use cases in the tests for mu/mpc.

usecases/mu/mpc/config.go
sthelen-enqs: We added tests for this to mu/mpc/config_test.go

usecases/mu/mpc/usecase.go
sthelen-enqs: We split the constituent measurement types into separate methods, and adjusted AddFeatures and AddUsecase to return error for all usecases

usecases/mu/mpc/usecase_test.go
DerAndereAndi, line 1: This file contains some unused variables in `BasicSuite`. Would be great to remove or comment them.

features/server/electricalconnection.go
```diff
@@ -31,6 +31,48 @@ func NewElectricalConnection(localEntity spineapi.EntityLocalInterface) (*Electr
 	return ec, nil
 }
 
+func (e *ElectricalConnection) GetOrAddIdForDescription(
```
DerAndereAndi, line 39: Please add comments for this function to understand its usage

features/server/electricalconnection.go
```diff
@@ -31,6 +31,48 @@ func NewElectricalConnection(localEntity spineapi.EntityLocalInterface) (*Electr
...
+			if description.ElectricalConnectionId != nil &&
+				description.PowerSupplyType == electricalConnectionDescription.PowerSupplyType &&
+				description.AcConnectedPhases == electricalConnectionDescription.AcConnectedPhases &&
+				description.AcRmsPeriodDuration == electricalConnectionDescription.AcRmsPeriodDuration &&
+				description.PositiveEnergyDirection == electricalConnectionDescription.PositiveEnergyDirection &&
+				description.ScopeType == electricalConnectionDescription.ScopeType &&
+				description.Label == electricalConnectionDescription.Label &&
+				description.Description == electricalConnectionDescription.Description {
+				electricalConnectionId = description.ElectricalConnectionId
+				break
```
DerAndereAndi: How about returning here directly? That would remove the need for the if statement in line 62

usecases/cs/lpc/usecase.go
```diff
@@ -172,15 +173,18 @@ func (e *LPC) loadControlWriteCB(msg *spineapi.Message) {
...
+func (e *LPC) AddFeatures() error {
 	// client features
 	_ = e.LocalEntity.GetOrAddFeature(model.FeatureTypeTypeDeviceDiagnosis, model.RoleTypeClient)
 
 	// server features
 	f := e.LocalEntity.GetOrAddFeature(model.FeatureTypeTypeLoadControl, model.RoleTypeServer)
 	f.AddFunctionType(model.FunctionTypeLoadControlLimitDescriptionListData, true, false)
 	f.AddFunctionType(model.FunctionTypeLoadControlLimitListData, true, true)
-	_ = f.AddWriteApprovalCallback(e.loadControlWriteCB)
+	err := f.AddWriteApprovalCallback(e.loadControlWriteCB)
```
DerAndereAndi: How about (for consistency) use `if err := ...; err != nil ...` ?

usecases/cs/lpc/usecase.go
```diff
@@ -204,6 +208,8 @@ func (e *LPC) AddFeatures() {
 			},
 		}
 		_ = lc.UpdateLimitDataForIds(newLimiData)
+	} else {
```
DerAndereAndi: With these long if statements, it might be easier to read the code if the error and return are handled directly in the if statement.

So changing instances like this into 

```golang
dcs, err := ...
if err != nil { 
  return err
}
```

That also removes the need for an if statement for success.

In addition I think we should check very early for feature existence and exit before adding any data if any of the required features are missing. This applies to all modified usecase.go implementations.

service/service.go
```diff
@@ -193,11 +193,23 @@ func (s *Service) IsRunning() bool {
...
+//
+// errors should not occur during normal usage of eebus-go, and should
+// generally be considered fatal implementation errors
+//
+// see usecase.AddFeatures() for more information
+func (s *Service) AddUseCase(useCase api.UseCaseInterface) error {
 	s.usecases = append(s.usecases, useCase)
 
-	useCase.AddFeatures()
+	err := useCase.AddFeatures()
```
DerAndereAndi: How about changing this to `if err := ...; err != nil { ...` ?

service/service_test.go
```diff
@@ -63,7 +63,7 @@ func (s *ServiceSuite) BeforeTest(suiteName, testName string) {
 
 func (s *ServiceSuite) Test_AddUseCase() {
 	ucMock := mocks.NewUseCaseInterface(s.T())
-	ucMock.EXPECT().AddFeatures().Return().Once()
+	ucMock.EXPECT().AddFeatures().Return(nil).Once()
 	ucMock.EXPECT().AddUseCase().Return().Once()
 
 	s.sut.AddUseCase(ucMock)
```
DerAndereAndi, line 69: Please check for the error return value 

```
	err := s.sut.AddUseCase(ucMock)
	assert.Nil(s.T(), err)
```

and add handling of the error case to this test, e.g. via:

```
func (s *ServiceSuite) Test_AddUseCase_Error() {
	ucMock := mocks.NewUseCaseInterface(s.T())

	expErr := errors.New("error")
	ucMock.EXPECT().AddFeatures().Return(expErr).Once()

	err := s.sut.AddUseCase(ucMock)
	assert.NotNil(s.T(), err)
	assert.Equal(s.T(), expErr, err)
}
```

usecases/cs/lpc/usecase.go
DerAndereAndi, line 1: If would love to see added testcases for checking the error handling, e.g. by adding tests to usecase_test.go like:

```golang
func (s *CsLPCSuite) Test_AddFeatures() {
	localEntity := spinemocks.NewEntityLocalInterface(s.T())
	s.sut.LocalEntity = localEntity

	mockLc := spinemocks.NewFeatureLocalInterface(s.T())
	mockLc.EXPECT().AddFunctionType(mock.Anything, mock.Anything, mock.Anything).Return()

	localEntity.EXPECT().GetOrAddFeature(model.FeatureTypeTypeDeviceDiagnosis, model.RoleTypeClient).Return(nil).Once()
	localEntity.EXPECT().GetOrAddFeature(model.FeatureTypeTypeLoadControl, model.RoleTypeServer).Return(mockLc).Once()

	expErr := errors.New("error")
	mockLc.EXPECT().AddWriteApprovalCallback(mock.Anything).Return(expErr).Once()

	err := s.sut.AddFeatures()
	assert.NotNil(s.T(), err)
}
```

usecases/usecase/usecase.go
```diff
@@ -114,6 +117,10 @@ func (u *UseCaseBase) IsCompatibleEntityType(entity spineapi.EntityRemoteInterfa
 		return false
 	}
 
+	if u.allEntityTypesValid {
```
DerAndereAndi, line 120: Please add a test for this scenario, e..g by adding the following in usecase_test.go link 26:

```golang
	s.uc.allEntityTypesValid = true
	result = s.uc.IsCompatibleEntityType(payload.Entity)
	assert.True(s.T(), result)
```

usecases/mu/mpc/usecase.go
```diff
@@ -0,0 +1,560 @@
...
+
+	parameterDescriptionId := electricalConnection.AddParameterDescription(parameterDescription)
+	if parameterDescriptionId == nil {
+		return errors.New("could not add parameter description")
+	}
+
+	return nil
+}
+
+func (e *MPC) getMeasurementDataForId(id *model.MeasurementIdType) (float64, error) {
```
DerAndereAndi, line 514: Please add tests for this function implementation

usecases/mu/mpc/usecase.go
```diff
@@ -0,0 +1,560 @@
...
+
+	parameterDescription := model.ElectricalConnectionParameterDescriptionDataType{
+		ElectricalConnectionId: electricalConnectionId,
+		MeasurementId:          e.acFrequency,
+		VoltageType:            util.Ptr(model.ElectricalConnectionVoltageTypeTypeAc),
+	}
+
+	parameterDescriptionId := electricalConnection.AddParameterDescription(parameterDescription)
+	if parameterDescriptionId == nil {
+		return errors.New("could not add parameter description")
```
DerAndereAndi, line 508: These error cases could be tested by passing mocks of `server.ElectricalConnection` and return an error on the function call.

Applies to all the other private functions as well.

usecases/mu/mpc/public.go
```diff
@@ -0,0 +1,545 @@
...
+	measurementDataForIds := make([]api.MeasurementDataForID, 0)
+
+	for _, measurementDataForId := range updateData {
+		if !measurementDataForId.Supported() {
+			return measurementDataForId.NotSupportedError()
+		} else {
```
DerAndereAndi: The `else` statement is not needed here, as we return in the other case.

So this code could be simplified to:

```golang
if !measurementDataForId.Supported() {
  return measurementDataForId.NotSupportedError()
}

measurementDataForIds = append(measurementDataForIds, measurementDataForId.MeasurementData())
```

DerAndereAndi (changes requested):
We are getting there :)

usecases/mu/mpc/events.go
```diff
@@ -0,0 +1,10 @@
+package mpc
+
+import (
+	spineapi "github.com/enbility/spine-go/api"
+)
+
+// handle SPINE events
+func (e *MPC) HandleEvent(payload spineapi.EventPayload) {
+	// No events supported
```
sthelen-enqs:
```suggestion
	// No events expected as ...
```

expand

usecases/ma/mpc/usecase.go
msbakhshi, line 26: Should not the use case actor type be UseCaseActorTypeMonitoringAppliance?

sthelen-enqs: Similar to validEntityTypes, validActorTypes is used as a filter for the remote Actor.
This means that if you add the MaMpc use case to a localEntity, it will search for connected remote devices with a MonitoredUnit actor that it can subscribe to measurements on

msbakhshi: Okay I get it. I reviewed it with the thought that this is mu/mpc.

PR #110, opened 2024-09-05
Fix LoadControl for 1 phase connected EVSE

DerAndereAndi:
- If an EVSE is connected via 1 phase, `LoadControlLimits` should not add empty data for non existing phases, otherwise this will lead to invalid data for consumers. E.g. checking if any phase is active will always fail, as these are always returning false
- Add more tests for Audi Mobile Connect setup with 1 phase
- Add more tests for Elli Connect/Pro Gen 1 with 1 phase setup

### DIFF
--- a/features/internal/electricalconnection_test.go
+++ b/features/internal/electricalconnection_test.go
@@ -549,15 +549,67 @@ func (s *ElectricalConnectionSuite) Test_GetLimitsForParameterId() {
 	assert.Equal(s.T(), defaultV, 0.1)
 }
 
+func (s *ElectricalConnectionSuite) Test_GetLimitsForParameterId_Elli_1Phase() {
+	filter := model.ElectricalConnectionPermittedValueSetDataType{
+		ParameterId: util.Ptr(model.ElectricalConnectionParameterIdType(0)),
+	}
+	minV, maxV, defaultV, err := s.localSut.GetPermittedValueDataForFilter(filter)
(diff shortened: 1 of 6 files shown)


features/internal/electricalconnection_test.go
```diff
@@ -549,25 +549,77 @@ func (s *ElectricalConnectionSuite) Test_GetLimitsForParameterId() {
...
 }
 
+func (s *ElectricalConnectionSuite) Test_GetLimitsForParameterId_Elli_1Phase() {
+	filter := model.ElectricalConnectionPermittedValueSetDataType{
+		ParameterId: util.Ptr(model.ElectricalConnectionParameterIdType(0)),
+	}
+	minV, maxV, defaultV, err := s.localSut.GetPermittedValueDataForFilter(filter)
+	assert.NotNil(s.T(), err)
+	assert.Equal(s.T(), minV, 0.0)
+	assert.Equal(s.T(), maxV, 0.0)
```
andig, line 559: I think you could simplify as

    s.Equal(minV…)

DerAndereAndi: I like things to be explicit. But thanks for the hint.

andig: Imho that is explicit since the Suite already knows the testing.T. Fwiw, with require it‘s a bit different:

    s.Require().Equal(…)

PR #115, opened 2024-10-01
Allow to update multiple data points in one step

DerAndereAndi:
Setting multiple data points, e.g. measurements for multiple phases, or setting currents and power values, should be done in one SPINE update call. Otherwise every single update will trigger an individual SPINE notification which is not nice to have.

This change updates the LoadControl and Measurement server features to allow providing multiple data points with individual filters for each, in a single call. This will then update the storage only once and trigger only one update notification.


DerAndereAndi:
@sthelen-enqs this is a suggestion on how to solve the issue described in the description and useful for your MPC draft PR. What do you think?

features/server/loadcontrol.go
```diff
@@ -71,48 +71,73 @@
...
+	deleteId *model.LoadControlLimitIdType,
 	deleteElements *model.LoadControlLimitDataElementsType,
-	limitId model.LoadControlLimitIdType,
 ) (resultErr error) {
-	return l.UpdateLimitDataForFilter(data, deleteElements, model.LoadControlLimitDescriptionDataType{LimitId: &limitId})
+	var filterData []api.LoadControlLimitDataForFilter
+	for _, item := range data {
+		filterData = append(filterData, api.LoadControlLimitDataForFilter{
+			Data:   item.Data,
+			Filter: model.LoadControlLimitDescriptionDataType{LimitId: &item.Id},
```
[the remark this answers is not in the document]
DerAndereAndi: Fixed

features/server/measurement.go
```diff
@@ -73,48 +73,73 @@
...
+	deleteId *model.MeasurementIdType,
 	deleteElements *model.MeasurementDataElementsType,
-	measurementId model.MeasurementIdType,
 ) (resultErr error) {
-	return m.UpdateDataForFilter(data, deleteElements, model.MeasurementDescriptionDataType{MeasurementId: &measurementId})
+	var filterData []api.MeasurementDataForFilter
+	for _, item := range data {
+		filterData = append(filterData, api.MeasurementDataForFilter{
+			Data:   item.Data,
+			Filter: model.MeasurementDescriptionDataType{MeasurementId: &item.Id},
```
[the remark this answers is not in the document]
DerAndereAndi: Fixed

api/featuresserver.go
```diff
@@ -90,26 +100,38 @@ type LoadControlServerInterface interface {
...
-	// Elements provided in deleteElements will be removed from the data set before the update
+	// Id provided in deleteId will trigger removal of matching items from the data set before the update
+	// Elements provided in deleteElement will limit the fields to be removed using Id
 	//
 	// Will return an error if the data set could not be updated
-	UpdateLimitDataForId(
-		data model.LoadControlLimitDataType,
+	UpdateLimitDataForIds(
+		data []LoadControlLimitDataForID,
+		deleteId *model.LoadControlLimitIdType,
```
DAMEK86: Why the Id is called `deleteId` when u have a method to update the limit?
It feels like the method name and the id doesn't match to each other

DerAndereAndi: On each update, you can also request to delete some data. So the `data` structure contains everything you want to set, and `deleteId` and `deleteElements` references what you want to delete.

DAMEK86: But shouldn't that be an delete instead of an update method?
It's confusing why I need to pass a delete ID for an CRUD method named update..

DerAndereAndi: It's to allow deletion of a field item, e.g. duration in LPC. The `data` structure doesn't allow you that. Having separate methods would requiring to introduce transactions, which I want to avoid.

The reasoning behind all that is how EEBUS works. See specs about partial updates.

DerAndereAndi: One addition: the complete updates are always partial and only one delete element can be done on a single SPINE message. That's why the `deleteId` and `deleteElements` are not part of the `data` structure.

sthelen-enqs:
> @sthelen-enqs this is a suggestion on how to solve the issue described in the description and useful for your MPC draft PR. What do you think?

Looks decent, I can try using it in the MPC draft PR and seeing how it works out in practice.

PR #123, opened 2024-10-15
MDT Support

heavyweight87:
MDT tested with a Vailant heatpump


DerAndereAndi:
Thanks a lot for this PR! The changes and additions look good so far. Would it be possible to get a SPINE trace log to see the communication in action? That would be the best source for me to evaluate the actual implementation.

Additionally I started adding comments to the `New...` methods in all usecases. Would be great if you could add that as well, see https://github.com/enbility/eebus-go/pull/136

heavyweight87:
Sure ill try to get these for you
We actually now parse the trace logs and save the json payloads to a file so its easier to debug issues

Any thoughts about adding something like this to the lower comms level? A flag that can be set to allow saving rx/tx packets to a file for logging?

DerAndereAndi:
I am not sure I understand your question. Could you elaborate a bit more on that please? (You can also join the Slack channel for such discussions, linked in the enbility.net contact page)

heavyweight87:
A feature that would allow saving of the incoming and outgoing packets to a file, for later inspection. It would be great for debugging purposes especially as due to the encryption tcpdump is pretty useless

DerAndereAndi:
That's imho already possible by implementing the logging interface and writing that output to a file. See https://github.com/enbility/eebus-go/blob/dev/examples/hems/main.go#L96 and https://github.com/enbility/eebus-go/blob/dev/examples/hems/main.go#L322-L368

heavyweight87:
[websocketLogs.json](https://github.com/user-attachments/files/17729835/websocketLogs.json)

Hi,
I attached a log - we store it in json format so its easier to read

There are other usecases here as well (new ones me and my colleague david opened) so there may be a bit of noise

usecases/api/ma_mdt.go
```diff
@@ -0,0 +1,24 @@
...
+
+	// return the momentary temperature of the domestic hot water circuit
+	//
+	// parameters:
+	//   - entity: the entity of the device (e.g. DHWCircuit)
+	//
+	// possible errors:
+	//   - ErrDataNotAvailable if no such limit is (yet) available
+	//   - and others
+	Temperature(entity spineapi.EntityRemoteInterface) (float64, error)
```
Ka0o0, line 23: According to the Use Case specification, every reading of a DHW temperature can be in a different state: normal, out of range, erroneous. We have to think about how we can handle this.

Ka0o0, line 23: We also need to think about how we want to handle different units. According to the UC spec, the measurement can be reported in different units. I talked to Andreas and we said it makes sense to let the user request the temperatures in different units, so this interface needs to be adapted.

sthelen-enqs:
Closed due to lack of activity. We may pick this up with changes or a fresh implementation at some point.

PR #129, opened 2024-10-28
JSON RPC2 implementation

DerAndereAndi:
This is an initial commit to start development of this feature in its own feature branch.

Please provide PRs against this branch when adding functionalities. The tasks are being coordinated in https://github.com/enbility/eebus-go/issues/130


DerAndereAndi:
If there are no more imminent changes planned, I would like to merge this into the dev branch. We should think of moving the JSON RPC2 interface into its own repo and work on an API design for that as well. What do you think?

sthelen-enqs:
The only imminent change I had planned would have been dropping the custom serialization of DeviceRemote/EntityRemote, but if I wanted to do that cleanly I'd also have to take a stab at the proposed case normalization in input/output data types. It might make more sense to work on that after this has been split into its own repo as part of the API design.

I'm fine with merging this and preparing to split this into its own repo and then moving from there.

PR #131, opened 2024-10-29
Forward error types as their string representation

sthelen-enqs:
By default go serializes `error` types as an empty struct `{}`, this patch would adjust that so that errors are serialized as their string representation by calling `Error()` on any non-nil types that implement the error interface.

This should be an improvement over the current state as it now allows identifying the cause of an error from the RPC client. An alternative could be to serialize the error as a struct of `{"type": <>, "message": <>}` using reflection to get the type name (and package path) of the error though I'm not sure if it's necessary and I'd have to check how easy that would be to get working.

### DIFF
--- a/examples/remote/util.go
+++ b/examples/remote/util.go
@@ -25,11 +25,23 @@ func isExportedOrBuiltinType(t reflect.Type) bool {
 	return token.IsExported(t.Name()) || t.PkgPath() == ""
 }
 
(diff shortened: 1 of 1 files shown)


DerAndereAndi:
I think providing the error as a string representation is just fine. In the stack the error is always defined as a string and a type is never used.

sthelen-enqs:
In that case I'll mark this as ready for review unless anyone else has any comments

PR #133, opened 2024-10-31
Fix Docs in EG/LPP types.go

anjrew:
Fix incorrect code comments


DerAndereAndi:
Hi, thank you for your PRs. Since they are all very small and have similar changes, could you please make one PR instead of 3?

I'd also like to hint at the content of CONTRIBUTION document.

Thank you again

PR #137, opened 2024-11-05
 Fix inconsistent DataUpdate event type identifiers for LPP use cases

anjrew:
This pull request resolves inconsistencies in the event type identifiers within usecases/cs/lpp/types.go and usecases/eg/lpp/types.go. The following changes were made:

 - Corrected event type identifiers to align with LPP scenarios instead of LPC scenarios where applicable. This helps ensure the identifiers accurately reflect their use cases and scenarios.
- Updated DataUpdateHeartbeat in both files to reflect the correct uclppserver and cs-lpp prefixes.

These changes improve code accuracy, documentation and maintain the correct mapping of constants to their respective use cases in the EEBUS library.

### DIFF
--- a/usecases/cs/lpp/types.go
+++ b/usecases/cs/lpp/types.go
@@ -43,5 +43,5 @@ const (
 	// E.g. going into or out of the Failsafe state
 	//
 	// Use Case LPP, Scenario 3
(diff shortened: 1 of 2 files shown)


DerAndereAndi:
Thanks you very much @anjrew ! Much appreciated.

PR #139, opened 2024-11-06
Move and update api.ManufacturerData

DerAndereAndi:
- move api.ManufacturerData to usecases/api
- fix serialization not to use SPINE schema


randomdudebunchofnumbers:
Works like a charm. Thank you very much.

PR #141, opened 2024-11-11
Make DataUpdateHeartbeat event names more consistent

sthelen-enqs:
This also allows differentiating between eglpc.DataUpdateHeartbeat and cslpc.DataUpdateHeartbeat events.

While syncing my branch on dev, I saw https://github.com/enbility/eebus-go/pull/137 and noticed that the DataUpdateHeartbeat events for eg/lpc and cs/lpc used the same identifier which means they can't be differentiated from another over e.g. json-rpc. I also took the liberty of changing the "uclppserver-" identifier to "cs-lpp-" to make it consistent with the other event names.

### DIFF
--- a/usecases/cs/lpp/types.go
+++ b/usecases/cs/lpp/types.go
@@ -43,5 +43,5 @@ const (
 	// E.g. going into or out of the Failsafe state
 	//
 	// Use Case LPP, Scenario 3
-	DataUpdateHeartbeat api.EventType = "uclppserver-DataUpdateHeartbeat"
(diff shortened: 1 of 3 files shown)


DerAndereAndi:
Thank you Simon. How about adding this into https://github.com/enbility/eebus-go/pull/136 instead of dev ?

sthelen-enqs:
That would work for me, but we might need to sync #136 from master because this PR bases on changes from #137 which were merged onto dev already.

DerAndereAndi:
Updated the maintenance branch

PR #144, opened 2024-11-26
Fix: have CurrentLimits method in opev and oscev filter for LoadControlLimitDescription instead of MeasurementDescription

AhmedMagdyCoretech:
Since the "ScopeType" field in the LoadControlLimitDescription is more relevant (with a value of "overloadProtection" and "selfConsumption" for OPEV and OSCEV, respectively), it shall be used as the filter for obtaining the relevant ElectricalConnectionPermittedValueSetListData items in the CurrentLimits getters for [OPEV](https://github.com/enbility/eebus-go/blob/59e7016a7018364d209f72331685b0ef0ba97b67/usecases/cem/opev/public.go#L19) and [OSCEV](https://github.com/enbility/eebus-go/blob/59e7016a7018364d209f72331685b0ef0ba97b67/usecases/cem/oscev/public.go#L19). From the relevant LoadControlLimitDescriptionData items, one can get their corresponding measurement IDs and use them to filter for relevant parameter description items.

### DIFF
--- a/usecases/cem/opev/public.go
+++ b/usecases/cem/opev/public.go
@@ -47,14 +47,18 @@ func (e *OPEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64, 
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	if len(limitDescs) != 3 {
+	if len(limitDescs) == 0 {
 		return nil, nil, nil, api.ErrDataNotAvailable
 	}
 
 	measDescs := make([]model.MeasurementDescriptionDataType, 0)
 	for _, ld := range limitDescs {
+		measId := ld.MeasurementId
+		if measId == nil {
+			return nil, nil, nil, api.ErrDataNotAvailable
+		}
 		filter := model.MeasurementDescriptionDataType{
-			MeasurementId: ld.MeasurementId,
+			MeasurementId: measId,
 		}
 		mds, err := meas.GetDescriptionsForFilter(filter)
 		if err != nil {
--- a/usecases/cem/opev/public.go
+++ b/usecases/cem/opev/public.go
@@ -26,23 +26,46 @@ func (e *OPEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64, 
 		return nil, nil, nil, err
 	}
 
+	lc, err := client.NewLoadControl(e.LocalEntity, entity)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
 	meas, err := client.NewMeasurement(e.LocalEntity, entity)
 	if err != nil {
 		return nil, nil, nil, err
 	}
 
-	filter := model.MeasurementDescriptionDataType{
-		MeasurementType: util.Ptr(model.MeasurementTypeTypeCurrent),
-		CommodityType:   util.Ptr(model.CommodityTypeTypeElectricity),
-		Unit:            util.Ptr(model.UnitOfMeasurementTypeA),
-		ScopeType:       util.Ptr(model.ScopeTypeTypeACCurrent),
+	filter := model.LoadControlLimitDescriptionDataType {
+		LimitType:     util.Ptr(model.LoadControlLimitTypeTypeMaxValueLimit),
+		LimitCategory: util.Ptr(model.LoadControlCategoryTypeObligation),
+		Unit:          util.Ptr(model.UnitOfMeasurementTypeA),
+		ScopeType:     util.Ptr(model.ScopeTypeTypeOverloadProtection),
 	}
-	measDesc, err := meas.GetDescriptionsForFilter(filter)
+	
+	limitDescs, err := lc.GetLimitDescriptionsForFilter(filter)
 	if err != nil {
 		return nil, nil, nil, err
 	}
+	if len(limitDescs) != 3 {
+		return nil, nil, nil, api.ErrDataNotAvailable
+	}
 
-	return ec.GetPhaseCurrentLimits(measDesc)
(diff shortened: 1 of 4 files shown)


DerAndereAndi:
Thank you @AhmedMagdyCoretech , this looks great!

usecases/cem/opev/public.go
```diff
@@ -26,23 +26,46 @@ func (e *OPEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64,
...
 	}
+	if len(limitDescs) != 3 {
+		return nil, nil, nil, api.ErrDataNotAvailable
+	}
 
-	return ec.GetPhaseCurrentLimits(measDesc)
+	measDescs := make([]model.MeasurementDescriptionDataType, 0)
+	for _, ld := range limitDescs {
+		filter := model.MeasurementDescriptionDataType{
+			MeasurementId: ld.MeasurementId,
```
DerAndereAndi: This assumes `ld.MeasurementId` is never nil, but there are devices that provide loadControlLimits where all fields but the ID are empty.

We should check if all limits do provide a non nil `MeasurementId` and only then continue, otherwise return an error. This check could be added in line 50.

usecases/cem/opev/public.go
```diff
@@ -26,23 +26,46 @@ func (e *OPEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64,
...
+		Unit:          util.Ptr(model.UnitOfMeasurementTypeA),
+		ScopeType:     util.Ptr(model.ScopeTypeTypeOverloadProtection),
 	}
-	measDesc, err := meas.GetDescriptionsForFilter(filter)
+	
+	limitDescs, err := lc.GetLimitDescriptionsForFilter(filter)
 	if err != nil {
 		return nil, nil, nil, err
 	}
+	if len(limitDescs) != 3 {
```
DerAndereAndi: This check is not correct, as wallboxes could be connected with 1 Phase only. So we should only check for the array not to be empty.

usecases/cem/oscev/public.go
```diff
@@ -26,23 +26,46 @@ func (e *OSCEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64,
...
+		Unit:          util.Ptr(model.UnitOfMeasurementTypeA),
+		ScopeType:     util.Ptr(model.ScopeTypeTypeSelfConsumption),
 	}
-	measDesc, err := meas.GetDescriptionsForFilter(filter)
+	
+	limitDescs, err := lc.GetLimitDescriptionsForFilter(filter)
 	if err != nil {
 		return nil, nil, nil, err
 	}
+	if len(limitDescs) != 3 {
```
DerAndereAndi: Same applies as in OPEV

usecases/cem/oscev/public.go
```diff
@@ -26,23 +26,46 @@ func (e *OSCEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64,
...
+		return nil, nil, nil, api.ErrDataNotAvailable
+	}
 
-	return ec.GetPhaseCurrentLimits(measDesc)
+	measDescs := make([]model.MeasurementDescriptionDataType, 0)
+	for _, ld := range limitDescs {
+		filter := model.MeasurementDescriptionDataType{
+			MeasurementId: ld.MeasurementId,
```
DerAndereAndi: Same applies as in OPEV

DerAndereAndi (changes requested):
Looks great overall, just some minor adjustments needed.

AhmedMagdyCoretech:
> Looks great overall, just some minor adjustments needed.

Thank you for the feedback. All the issues have been duly resolved.

usecases/cem/opev/public.go
```diff
@@ -26,23 +26,38 @@ func (e *OPEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64,
...
+	}
 
-	return ec.GetPhaseCurrentLimits(measDesc)
+	measDescs := make([]model.MeasurementDescriptionDataType, 0)
+	for _, ld := range limitDescs {
+		measId := ld.MeasurementId
+		if measId == nil {
+			return nil, nil, nil, api.ErrDataNotAvailable
+		}
+		filter := model.MeasurementDescriptionDataType{
```
DerAndereAndi: The variable name confuses me a bit. Maybe just use `measDesc` instead?

usecases/cem/oscev/public.go
```diff
@@ -26,23 +26,38 @@ func (e *OSCEV) CurrentLimits(entity spineapi.EntityRemoteInterface) ([]float64,
...
+	}
 
-	return ec.GetPhaseCurrentLimits(measDesc)
+	measDescs := make([]model.MeasurementDescriptionDataType, 0)
+	for _, ld := range limitDescs {
+		measId := ld.MeasurementId
+		if measId == nil {
+			return nil, nil, nil, api.ErrDataNotAvailable
+		}
+		filter := model.MeasurementDescriptionDataType{
```
DerAndereAndi: Same as in the other usecase

DerAndereAndi (changes requested):
I like the changes, just a minor last request regarding a variable name. Sorry if this looks picky, but I understand something else with `filters` .

AhmedMagdyCoretech:
> I like the changes, just a minor last request regarding a variable name. Sorry if this looks picky, but I understand something else with `filters` .

Thank you for the feedback. Not so picky, I would say. Aptly addressed.

DerAndereAndi (approved):

DerAndereAndi:
Thank you very much @AhmedMagdyCoretech !

PR #150, opened 2025-01-07
Add write approval callbacks for Failsafe values of CS-LPC usecase

sthelen-enqs:
This is a work-in-progress proposal for adding write approval callbacks for the failsafe values of the CS-LPC usecase.

Code changes should be functional, but are not yet final. Cleanup and thorough testing may still be required.

Fixes/updates from our side should come when I can find the time to do so. If someone wants to use/update this in the meantime, feel free.

Comments/fixes etc are welcome.

### DIFF
--- a/examples/hems/main.go
+++ b/examples/hems/main.go
@@ -164,11 +164,9 @@ func (h *hems) OnLPCEvent(ski string, device spineapi.DeviceRemoteInterface, ent
 		for msgCounter, configs := range pendingDeviceConfigWrites {
 			fmt.Printf("Approving LPC device config write with msgCounter %d for features: ", msgCounter)
 			for _, config := range configs {
-				keyName := "nil"
 				if config.Description.KeyName != nil {
-					keyName = string(*config.Description.KeyName)
+					fmt.Printf("%s ", string(*config.Description.KeyName))
 				}
-				fmt.Printf("%s ", keyName)
 			}
 			fmt.Print("\n")
 			h.uccslpc.ApproveOrDenyDeviceConfiguration(msgCounter, true, "")
@@ -201,11 +199,9 @@ func (h *hems) OnLPPEvent(ski string, device spineapi.DeviceRemoteInterface, ent
 		for msgCounter, configs := range pendingDeviceConfigWrites {
 			fmt.Printf("Approving LPP device config write with msgCounter %d for features: ", msgCounter)
 			for _, config := range configs {
-				keyName := "nil"
 				if config.Description.KeyName != nil {
-					keyName = string(*config.Description.KeyName)
+					fmt.Printf("%s ", string(*config.Description.KeyName))
 				}
-				fmt.Printf("%s ", keyName)
 			}
 			fmt.Print("\n")
 			h.uccslpp.ApproveOrDenyDeviceConfiguration(msgCounter, true, "")
--- a/examples/hems/main.go
+++ b/examples/hems/main.go
@@ -164,9 +164,7 @@ func (h *hems) OnLPCEvent(ski string, device spineapi.DeviceRemoteInterface, ent
 		for msgCounter, configs := range pendingDeviceConfigWrites {
 			fmt.Printf("Approving LPC device config write with msgCounter %d for features: ", msgCounter)
 			for _, config := range configs {
-				if config.Description.KeyName != nil {
-					fmt.Printf("%s ", string(*config.Description.KeyName))
-				}
+				fmt.Printf("%s ", string(config.KeyName))
 			}
 			fmt.Print("\n")
 			h.uccslpc.ApproveOrDenyDeviceConfiguration(msgCounter, true, "")
--- a/examples/hems/main.go
+++ b/examples/hems/main.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/enbility/eebus-go/api"
+	"github.com/enbility/eebus-go/features/server"
 	"github.com/enbility/eebus-go/service"
 	ucapi "github.com/enbility/eebus-go/usecases/api"
 	"github.com/enbility/eebus-go/usecases/cem/vabd"
@@ -159,13 +160,23 @@ func (h *hems) OnLPCEvent(ski string, device spineapi.DeviceRemoteInterface, ent
 			fmt.Println("Approving LPC limit write with msgCounter", msgCounter, "and limit", write.Value, "W")
 			h.uccslpc.ApproveOrDenyConsumptionLimit(msgCounter, true, "")
 		}
-		for msgCounter, config := range pendingDeviceConfigWrites {
-			fmt.Print("Approving LPC device config write with msgCounter ", msgCounter)
-			if config.FailsafeDuration != nil {
-				fmt.Printf(" and FailsafeDurationMinimum %s", *config.FailsafeDuration)
+		for msgCounter, configs := range pendingDeviceConfigWrites {
+			localEntity := h.myService.LocalDevice().EntityForType(model.EntityTypeTypeCEM)
+			dc, err := server.NewDeviceConfiguration(localEntity)
+			if err != nil {
+				fmt.Println("Not approving LPC device configuration writes because of error:")
+				fmt.Println(err)
+				return
 			}
-			if config.FailsafeLimit != nil {
-				fmt.Printf(" and FailsafeConsumptionLimit %f", *config.FailsafeLimit)
+			fmt.Printf("Approving LPC device config write with msgCounter %d ", msgCounter)
+			for _, config := range(configs) {
+				description, err := dc.GetKeyValueDescriptionFoKeyId(*config.KeyId)
+				if description == nil || err != nil {
+					fmt.Printf("LPC approving device configuation writes: no device configuration for KeyID %d found\n", *config.KeyId)
+					continue
+				}
+
+				fmt.Printf("including %s ", *description.KeyName)
 			}
 			fmt.Print("\n")
 			h.uccslpc.ApproveOrDenyDeviceConfiguration(msgCounter, true, "")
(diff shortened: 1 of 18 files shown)


DerAndereAndi:
Thanks for starting the work on this. I do now think that the write approval callback API should be used for all writes on any server feature.

sthelen-enqs:
I was waiting for some final qualification tests to ensure that a "spec-compliant" implementation can be achieved for LPC/LPP with this PR. The final test is still outstanding due to delays with the qualification, but I'm reasonably confident that no further changes are necessary to this PR to fully pass the qualification.

DerAndereAndi:
@sthelen-enqs I analyzed the PR with Claude and iterated a bit over it. Here is its summary:

Based on my analysis, the PR implementation is well-designed and follows the established patterns correctly. Here are my recommendations:

  1. Fix the Logging Issue

  In deviceConfigurationWriteCB, fix the format string:

```
  // Current (line 228)
  logging.Log().Debug("LPC deviceConfigurationWriteCB: no device configuration for KeyID %d found")

  // Should be:
  logging.Log().Debug("LPC deviceConfigurationWriteCB: no device configuration for KeyID found: ", *deviceKeyValueData.KeyId)
```

  2. Add Documentation

  Add a comment in the interface or README explaining the approval pattern:

  // WriteApprovalRequired event is fired when any server feature receives a write request
  // that requires approval. The application should:
  // 1. Check PendingConsumptionLimits() for LoadControl writes
  // 2. Check PendingDeviceConfigurations() for DeviceConfiguration writes
  // 3. Call the appropriate ApproveOrDeny* method with the msgCounter

  3. Consider Adding Helper Methods (Optional)

  For better developer experience, consider adding helper methods to check if specific configurations are pending:

```
  // Check if a specific configuration key is pending approval
  func (e *LPC) IsPendingDeviceConfiguration(msgCounter model.MsgCounterType, keyName model.DeviceConfigurationKeyNameType) bool {
      configs := e.PendingDeviceConfigurations()
      if pendingConfigs, exists := configs[msgCounter]; exists {
          for _, config := range pendingConfigs {
              if config.Description != nil && config.Description.KeyName != nil &&
                 *config.Description.KeyName == keyName {
                  return true
              }
          }
      }
      return false
  }
```

  4. Test Coverage

  Ensure tests cover:
  - Approval flow for both parameters
  - Auto-approval of unrelated DeviceConfiguration writes
  - Handling multiple configuration changes in one message
  - Thread safety of concurrent approvals
  - Timeout behavior (inherited from spine-go)

  Summary

  The PR implementation is functionally complete and well-designed. It correctly:
  - Follows the established LoadControl pattern
  - Provides selective filtering for relevant parameters
  - Maintains thread safety
  - Gives applications full control over approval decisions
  - Auto-approves unrelated writes to avoid blocking other use cases

  The only required change is the logging fix. The other suggestions are optional improvements for developer experience and documentation.

sthelen-enqs:
1. The logging issue should be fixed now
2. I updated the documentation in events.go for LPC/LPP with the new methods as well.
3. I don't think point 3 is particularly relevant since you need to check every pending device configuration (and then switch on the type) in the handler, I wouldn't recommend selectively handling device configurations.
4. I'll double-check the test coverage for this

usecases/cs/lpc/usecase.go
```diff
@@ -172,6 +177,80 @@ func (e *LPC) loadControlWriteCB(msg *spineapi.Message) {
...
+	}
+	for _, deviceKeyValueData := range data.DeviceConfigurationKeyValueData {
+		description, err := dc.GetKeyValueDescriptionFoKeyId(*deviceKeyValueData.KeyId)
+		if description == nil || err != nil {
+			logging.Log().Debug("LPC deviceConfigurationWriteCB: no device configuration for KeyID found: ", *deviceKeyValueData.KeyId)
+			continue
+		}
+
+		// Only ask for write approval if at least one of the configurations we care about is trying to be set
+		if _, exists := configsToApprove[*description.KeyName]; exists {
```
DerAndereAndi: Doesn't this panic when KeyName is nil?

usecases/cs/lpp/usecase.go
```diff
@@ -172,6 +177,80 @@ func (e *LPP) loadControlWriteCB(msg *spineapi.Message) {
...
+	}
+	for _, deviceKeyValueData := range data.DeviceConfigurationKeyValueData {
+		description, err := dc.GetKeyValueDescriptionFoKeyId(*deviceKeyValueData.KeyId)
+		if description == nil || err != nil {
+			logging.Log().Debug("LPP deviceConfigurationWriteCB: no device configuration for KeyID found: ", *deviceKeyValueData.KeyId)
+			continue
+		}
+
+		// Only ask for write approval if at least one of the configurations we care about is trying to be set
+		if _, exists := configsToApprove[*description.KeyName]; exists {
```
DerAndereAndi: Doesn't this panic when KeyName is nil?

examples/hems/main.go
```diff
@@ -152,12 +152,21 @@ func (h *hems) OnLPCEvent(ski string, device spineapi.DeviceRemoteInterface, ent
...
 		// approve any write
 		for msgCounter, write := range pendingWrites {
-			fmt.Println("Approving LPC write with msgCounter", msgCounter, "and limit", write.Value, "W")
+			fmt.Println("Approving LPC limit write with msgCounter", msgCounter, "and limit", write.Value, "W")
 			h.uccslpc.ApproveOrDenyConsumptionLimit(msgCounter, true, "")
 		}
+		for msgCounter, configs := range pendingDeviceConfigWrites {
+			fmt.Printf("Approving LPC device config write with msgCounter %d for features: ", msgCounter)
+			for _, config := range configs {
+				fmt.Printf("%s ", *config.Description.KeyName)
```
DerAndereAndi: What if Description or Keyname are nil?

usecases/cs/lpp/public.go
```diff
@@ -267,6 +267,60 @@ func (e *LPP) SetFailsafeDurationMinimum(duration time.Duration, changeable bool
 	return dc.UpdateKeyValueDataForFilter(data, nil, filter)
 }
 
+// return the currently pending incoming failsafe consumption limit writes
```
DerAndereAndi: `consumption` -> `production`

usecases/cs/lpc/usecase.go
```diff
@@ -172,6 +177,80 @@ func (e *LPC) loadControlWriteCB(msg *spineapi.Message) {
...
+func (e *LPC) deviceConfigurationWriteCB(msg *spineapi.Message) {
+	if msg.RequestHeader == nil || msg.RequestHeader.MsgCounter == nil ||
+		msg.Cmd.DeviceConfigurationKeyValueListData == nil {
+		logging.Log().Debug("LPC deviceConfigurationWriteCB: invalid message")
+		return
+	}
+
+	data := msg.Cmd.DeviceConfigurationKeyValueListData
+
+	if data == nil || data.DeviceConfigurationKeyValueData == nil || len(data.DeviceConfigurationKeyValueData) == 0 {
```
DerAndereAndi: data can't be nil as it is assigned to `msg.Cmd.DeviceConfigurationKeyValueListData` which in line 201 is checked not to be nil. So not wrong, but will never be the case anyway

usecases/api/types.go
```diff
@@ -166,3 +166,9 @@ type DurationSlotValue struct {
 	Duration time.Duration // Duration of this slot
 	Value    float64       // Energy Cost or Power Limit
 }
+
+type PendingDeviceConfiguration struct {
+	Description       *model.DeviceConfigurationKeyValueDescriptionDataType `json:"description,omitempty"`
+	Value             *model.DeviceConfigurationKeyValueValueType           `json:"value,omitempty"`
+	IsValueChangeable *bool                                                 `json:"isValueChangeable,omitempty" eebus:"writecheck"`
```
DerAndereAndi: Why did you add the `writecheck` tag attribute here?

DerAndereAndi (changes requested):
In addition to my comments, we are missing tests for the new code.

Additionally `WriteApprovalRequired` message is now used for limit write approvals and failsafe and the application doesn't know which and both could also happen around the same time. In this case I would propose introduce a new message so the application can handle it more easily.

The newmethods in `usecase.go` and `public.go` are nearly identical. We could move that code into shared functions in the `internal` package.

usecases/internal/deviceconfiguration.go
```diff
@@ -0,0 +1,91 @@
...
+				IsValueChangeable: configKeyValueData.IsValueChangeable,
+			}
+
+			if _, exists := result[msgCounter]; !exists {
+				result[msgCounter] = []ucapi.PendingDeviceConfiguration{pendingConfigData}
+			} else {
+				result[msgCounter] = append(result[msgCounter], pendingConfigData)
+			}
+		}
+	}
```
DerAndereAndi, line 89: How about changing this to:

```go
for msgCounter, msg := range pendingDeviceConfigs {
    if msg.Cmd.DeviceConfigurationKeyValueListData == nil {
        continue
    }
    for _, configKeyValueData := range msg.Cmd.DeviceConfigurationKeyValueListData.DeviceConfigurationKeyValueData {
        if configKeyValueData.KeyId == nil {
            continue
        }
        description, err := dc.GetKeyValueDescriptionFoKeyId(*configKeyValueData.KeyId)
        if err != nil || description == nil || description.KeyName == nil {
            continue
        }
        result[msgCounter] = append(result[msgCounter], ucapi.PendingDeviceConfiguration{
            Description:       description,
            Value:             configKeyValueData.Value,
            IsValueChangeable: configKeyValueData.IsValueChangeable,
        })
    }
}
```

This way the example code does not need to check for `if config.Description.Keyname != nil` and also `config.Description` being `nil` is also avoided.

So the hems example would only have this:

```go
for _, config := range configs {
    fmt.Printf("%s ", string(*config.Description.KeyName))
}
```

usecases/cs/lpc/usecase.go
```diff
@@ -175,6 +179,51 @@ func (e *LPC) loadControlWriteCB(msg *spineapi.Message) {
...
+			e.pendingDeviceConfigMux.Unlock()
+			e.EventCB(msg.DeviceRemote.Ski(), msg.DeviceRemote, msg.EntityRemote, ConfigurationWriteApprovalRequired)
+			return
+		}
+		e.pendingDeviceConfigMux.Unlock()
+	}
+
+	// If neither a failsafe duration nor a failsafe limit were set this message does not pertain to this use case so we accept
+	go e.approveOrDenyDeviceConfiguration(msg, true, "")
+}
```
DerAndereAndi, line 225: `if approvalRequired==true` and the msgCounter is already in `pendingDeviceConfigs`, the code unlocks and then auto-approves the duplicate — approving a message whose original is still pending and leaving the original entry orphaned. This issue is pre-existing, but worth checking if we should improve this maybe?

usecases/cs/lpc/public.go
```diff
@@ -270,6 +271,31 @@ func (e *LPC) SetFailsafeDurationMinimum(duration time.Duration, changeable bool
 	return dc.UpdateKeyValueDataForFilter(data, nil, filter)
 }
 
+// return the currently pending incoming failsafe consumption limit writes
```
DerAndereAndi, line 274: Shouldn't this rather be:

```
// return the currently pending incoming device configuration writes
```

usecases/cs/lpp/public.go
```diff
@@ -272,6 +273,32 @@ func (e *LPP) SetFailsafeDurationMinimum(duration time.Duration, changeable bool
 	return dc.UpdateKeyValueDataForFilter(data, nil, filter)
 }
 
+// return the currently pending incoming failsafe production limit writes
```
DerAndereAndi, line 276: Shouldn't this rather be:

```
// return the currently pending incoming device configuration writes
```

DerAndereAndi (changes requested):
An additional note regarding LPC and LPP `PendingDeviceConfigurations()`: if LPC and LPP are on the same entity, both need to be approved, even though it is the same data point! The application has to be aware of this and we may want to think of adding tests for this?
We also should to re-check if spine-go handles this correctly.

PR #154, opened 2025-02-14
Report per phase measurements for correct phase

sthelen-enqs:
If/when a device registers PhaseSpecific measurements only for one specific phase (or registers them for all phases but only provides measurements for one of them), MeasurementPhaseSpecificDataForFilter() returns a []float64 containing only 1 value with no way for the API user to know for which phase the measurement was meant. This is an issue in e.g. ma/mpc's PowerPerPhase which would then return an array with less than 3 members and callers would then not know to which phase the measurement corresponds.

This PR changes the return type of MeasurementPhaseSpecificDataForFilter to always contain one entry per phase so that for a device which only reports data on phase B, it would return []float64{0, 10, 0} instead of []float64{10}. The advantage of this approach is that it doesn't change the return type of MeasurementPhaseSpecificDataForFilter or any of its callers, the downside is that callers can't (easily) distinguish devices that are reporting 0 for a phase and devices that aren't reporting a phase at all. The alternative would be to change the return type from []float64 to map[ElectricalConnectionPhaseNameType]float64 and then going through all the callers to make them use/return that as well.

I'm not particular to either approach so I picked the one that changed the surface API less.

### DIFF
--- a/usecases/internal/measurement.go
+++ b/usecases/internal/measurement.go
@@ -29,20 +29,31 @@ func MeasurementPhaseSpecificDataForFilter(
 	}
 
 	var result []float64
+	if validPhaseNameTypes != nil {
+		// pre-allocate result array for each possible phase so we can add phases to it in arbitrary order
+		result = make([]float64, len(validPhaseNameTypes))
+	}
 
 	for _, item := range data {
 		if item.Value == nil || item.MeasurementId == nil {
 			continue
 		}
 
+		phaseIndex := -1
 		if validPhaseNameTypes != nil {
 			filter := model.ElectricalConnectionParameterDescriptionDataType{
 				MeasurementId: item.MeasurementId,
 			}
 			param, err := electricalConnection.GetParameterDescriptionsForFilter(filter)
-			if err != nil || len(param) == 0 ||
-				param[0].AcMeasuredPhases == nil ||
-				!slices.Contains(validPhaseNameTypes, *param[0].AcMeasuredPhases) {
+			if err != nil || len(param) == 0 || param[0].AcMeasuredPhases == nil {
+				// error getting parameter description
+				continue
+			}
+
+			// calculate the offset into result for the measured phase
+			phaseIndex = slices.Index(validPhaseNameTypes, *param[0].AcMeasuredPhases)
+			if phaseIndex == -1 {
+				// ignore phase measurements not specified in validPhaseNameTypes
 				continue
 			}
 		}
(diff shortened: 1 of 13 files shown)


sthelen-enqs:
After a discussion, we'll rewrite this to return the map[ElectricalConnectionPhaseNameType]float64 to be more explicit about whether phases were written to or not.

andig:
Another and easier to use option might be a `[3]*float`.

sthelen-enqs:
I've updated this PR to return a map[ElectricalConnectionPhaseNameType]float64.
There is one incidental change which may be of interest. The acPowerTotal measurements must now have an acMeasuredPhases electricalConnectionParameterDescriptionData set for the MA MPC/MGCP usecases to pick up/report the value via the Power() functions. These are required values in the MPC/MGCP specification and should therefore not affect communication with real devices (in theory), but I had to adjust two test cases to add the required parameter.

sthelen-enqs:
We should update other usecases with the new approach to passing/receiving phase-specific values. That task is tracked in #197

usecases/internal/measurement.go
```diff
@@ -33,23 +28,37 @@ func MeasurementPhaseSpecificDataForFilter(
...
+			// error getting parameter description
+			continue
+		}
+
+		var phaseName model.ElectricalConnectionPhaseNameType
+		if param[0].AcMeasuredPhases != nil {
+			phaseName = *param[0].AcMeasuredPhases
+		} else if validPhaseNameTypes == nil {
+			// if we're not filtering by valid phase names, allow acMeasuredPhases to be unset
+			phaseName = model.ElectricalConnectionPhaseNameTypeNone
```
DerAndereAndi, line 52: Do we have a test for this code path?

usecases/ma/mgcp/public.go
```diff
@@ -78,7 +78,13 @@ func (e *MGCP) Power(entity spineapi.EntityRemoteInterface) (float64, error) {
 		return 0, api.ErrDataNotAvailable
 	}
 
-	return data[0], nil
+	for _, k := range data {
+		// If the Monitored Unit is connected to less than three phases, one of the other combinations like "a" or "ab" are allowed instead of "abc".
+		// The values "a", "b", and "c" are permitted if and only if only one
```
DerAndereAndi, line 83: the comment looks incomplete "... and only if only one ..." what? double only? :)

usecases/internal/measurement.go
```diff
@@ -33,23 +28,37 @@ func MeasurementPhaseSpecificDataForFilter(
...
+		} else {
+			// error getting parameter description
```
DerAndereAndi, line 54: I don't think this comment is correct here

DerAndereAndi (changes requested):
EVCEM is not updated but requires the same changes. Would be great if it would be part of this PR as well

PR #155, opened 2025-02-18
Small bugfixes for controlbox and hems example

heinemannj:
Very small bugfix in the LPP Event Handler of the hems example.
Bugfix in the LPP Event Handler of controlbox example. Don't use lpc interfaces in lpp scenarios.


DerAndereAndi:
Thank you very much for this findings and fixes, looks great!

PR #169, opened 2025-03-28
MPC and MGCP use cases as actors Monitored Unit and Grid Connection Point

jaroslawsky:
For both use cases, there has only been code available so far for the actor "Monitoring Appliance." This feature branch is now intended to also enable the implementation of the respective counterparts. Admittedly, the test cases are still missing here. If these are considered indispensable, please let me know. I will then endeavor to create them.

These implementations can be interactively tested using the controlbox simulator from this [repository](https://github.com/vollautomat/eebus-go/tree/simulators). Both the newly added "smartmeter" example and the modified "hems" example from the mentioned simulator repository can be used.

Both examples send a PowerLimitationFactor value one second after the connection is established and changing values for all other parameters every three seconds. These values are not necessarily very realistic. The aim was simply to send something that constantly changes. This makes it easiest to observe the functioning of the two use cases.

Using the simulator looks like this:
Bildschirmfoto 2025-03-28 um 10 05 37

### DIFF
--- a/examples/smartmeter/main.go
+++ b/examples/smartmeter/main.go
@@ -18,8 +18,6 @@ import (
 	"github.com/enbility/eebus-go/service"
 	ucapi "github.com/enbility/eebus-go/usecases/api"
 
-	// "github.com/enbility/eebus-go/usecases/cem/vabd"
-	// "github.com/enbility/eebus-go/usecases/cem/vapd"
 	gcpmgcp "github.com/enbility/eebus-go/usecases/gcp/mgcp"
 	mumpc "github.com/enbility/eebus-go/usecases/mu/mpc"
 	shipapi "github.com/enbility/ship-go/api"
@@ -35,9 +33,6 @@ type smartmeter struct {
 
 	ucgcpmgcp ucapi.GcpMGCPInterface
 	ucmumpc   ucapi.MuMPCInterface
-
-	// uccemvabd ucapi.CemVABDInterface
-	// uccemvapd ucapi.CemVAPDInterface
 
 	gridPowerLimitFactor float64
 	gridPower            float64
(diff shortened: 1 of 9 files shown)


jaroslawsky:
By the way, the MPC use case has already been successfully tested with real hardware - specifically with a Vaillant VR921.

DerAndereAndi:
Hi Klaus,

thanks for the PR and congrats on your efforts for creating the control box simulator. Well done!

 I will take a closer look as soon as I find some time. Some quick notes:

- An MPC Implementation is already being worked on and in the final stages to be merged. See https://github.com/enbility/eebus-go/pull/108 
- I would suggest to remove the MPC implementation from the PR
- Every implementation needs to have test coverage, I will not merge if the coverage goes down > 0.5%
- I would take a look at the code and provide review comments as I go through it with questions or directly requests for changes (or discussion of those). But I can't promise a timeframe where I can go over it and most likely the whole process will take some time until I would call it ready for merge

I mainly developed this all alone. Used my own experience and best practices for establishing a code and comment style. It is far from perfect. I did not write them down formally, getting the stack to this stage was way too much work already. So I put the efforts where I see them being of best use. This also means that going over a PR will be a bit more effort that otherwise, as contributors can only prepare by looking at the currently implemented code as an example.

Hope this sounds fine for you.

Best,
Andreas

jaroslawsky:
Hello Andreas,

I don’t expect more than what you’ve mentioned. Everything is fine. I’ll just wait until you’ve had time to look into it and then I’ll try to address any potential comments. If the MPC use case is excluded because it has already been implemented elsewhere, that’s absolutely no problem. I would then only need to focus on the tests for the MGCP use case, which makes things easier for me. However, I don’t have much experience with Go and its environment yet, so I’ll need to familiarize myself with the test framework first. It will also take me a little time on my end.

Best regards,
Klaus

andig:
Has this been superseeded by https://github.com/enbility/eebus-go/pull/224 Now?

sthelen-enqs:
This has been superseded by https://github.com/enbility/eebus-go/pull/224
Thank you for the PR

PR #181, opened 2025-07-12
feat: unified data   validation framework for EEBUS use cases

DerAndereAndi:
## Summary

Introduces a generic validation framework and applies it to the EG LPC, MA MGCP, and MA MPC use cases, so data returned by the public API methods is verified against EEBUS spec requirements before it reaches consumers.

The bug-fix portion of the original PR (DeviceConfiguration operator-precedence fix + event-firing consistency across the affected handlers) has been split into **#215** and should land first.

### Framework (`usecases/internal/`)

- `validation.go` — generic `Validator[T]` / `BaseValidator[T]` with composable rules (`RequireField`, `RequireScaledNumber`, `ValidateRange`, `ValidateMinMax`, `ValidateEnum`, `ValidateSliceNotEmpty`, `CombineRules`, `ConditionalRule`, `ValidateCustom`).
- `measurement.go` — `MeasurementValidator` with measurement-specific rules (`RequireMeasurementId`, `RequireMeasurementValue`, `RequireValueType`, `RequireValueSource`, `RequireValueSourceMandatory`, `ValidateValueState`), plus `SkipValueState` implementing MGCP-003 ("values with state `error` or `outOfRange` SHALL be ignored").
- `MeasurementPhaseSpecificDataForFilter` now takes a validator and skips individual failing measurements instead of aborting the whole call.
- `VALIDATION.md` — developer guide with usage examples.

### Use-case validators

- **EG LPC** — `LoadControlLimitData`, `DeviceConfigurationKeyValue`, `ElectricalConnectionCharacteristic` validators; wired into `ConsumptionLimit`, `FailsafeConsumptionActivePowerLimit`, `ConsumptionNominalMax`.
- **MA MGCP** — scenario-specific validators for Power, Energy, Current, Voltage, Frequency (incl. MGCP-003 skip behavior).
- **MA MPC** — scenario-specific validators for Power, Energy, Current, Voltage, Frequency.

## Behavior changes

- MPC/MGCP: when no measurement passes validation, both now consistently return `api.ErrDataNotAvailable` (previously divergent between the two use cases).
- MPC: drops the 45–65 Hz frequency-range check — not part of the MPC spec.

## Notes

- No public API changes.
- Depends on **#215**.
- Earlier review feedback from @sthelen-enqs is still applicable to the framework/validator specifics and will be addressed in follow-up commits.


DerAndereAndi:
@sthelen-enqs this should also help for the scenario in your PR https://github.com/enbility/spine-go/pull/52

usecases/eg/lpc/public_test.go
```diff
@@ -417,3 +419,663 @@ func (s *EgLPCSuite) Test_PowerConsumptionNominalMax() {
...
+			},
+		},
+	}
+
+	rFeature := s.remoteDevice.FeatureByEntityTypeAndRole(s.monitoredEntity, model.FeatureTypeTypeLoadControl, model.RoleTypeServer)
+	_, fErr := rFeature.UpdateData(true, model.FunctionTypeLoadControlLimitDescriptionListData, descData, nil, nil)
+	assert.Nil(s.T(), fErr)
+
+	// Test 1: Missing LimitId - should return ErrDataInvalid when validator runs
+	// Note: Data must include Value to pass the `value.Value == nil` check in public.go:59
```
sthelen-enqs, line 444: This comment is incorrect, and confusing. GetLimitDataForId will return nil, api.ErrDataNotAvailable because the filter won't match any limits and len(result) will be 0. The check on L59 then short-circuits on err != nil and value.Value will never be checked.
```suggestion
```

usecases/eg/lpc/public_test.go
```diff
@@ -417,3 +419,663 @@ func (s *EgLPCSuite) Test_PowerConsumptionNominalMax() {
...
+	invalidLimitData := &model.LoadControlLimitListDataType{
+		LoadControlLimitData: []model.LoadControlLimitDataType{
+			{
+				// LimitId missing - this should trigger validation error
+				IsLimitChangeable: util.Ptr(true),
+				IsLimitActive:     util.Ptr(false),
+				Value:             model.NewScaledNumberType(6000), // Include value to pass initial check
```
sthelen-enqs, line 451:
```suggestion
				Value:             model.NewScaledNumberType(6000),
```

usecases/eg/lpc/public_test.go
```diff
@@ -417,3 +419,663 @@ func (s *EgLPCSuite) Test_PowerConsumptionNominalMax() {
...
+
+	_, fErr = rFeature.UpdateData(true, model.FunctionTypeDeviceConfigurationKeyValueListData, validKeyData, nil, nil)
+	assert.Nil(s.T(), fErr)
+
+	data, err = s.sut.FailsafeDurationMinimum(s.monitoredEntity)
+	assert.Nil(s.T(), err)
+	assert.Equal(s.T(), time.Duration(time.Hour*2), data)
+}
+
+func (s *EgLPCSuite) Test_ConsumptionNominalMax_ValidationErrors() {
```
sthelen-enqs, line 806: This test case is using the wrong CharacteristicType throughout because the "current" version uses the DeviceType to determine the correct CharacterisicType, but the DeviceType is never initialised in the test suite (and therefore nil).

Could you maybe rebase this PR on fix/use-correct-eg-lpc-lpp-characteristic-types ?

sthelen-enqs (changes requested):
Hi,

I haven't fully checked through the whole PR, but found a couple of minor nits in the tests that might be nice to fix. I'll check over the rest of the validation later.

general nit: could you run gofmt on the test cases as well?

usecases/eg/lpc/public.go
```diff
@@ -289,6 +303,12 @@ func (e *LPC) ConsumptionNominalMax(entity spineapi.EntityRemoteInterface) (floa
 		return 0, api.ErrDataNotAvailable
 	}
 
+	// Validate ElectricalConnectionCharacteristicData using EG LPC validator
+	// This ensures nominal consumption values are positive and within reasonable ranges
```
sthelen-enqs: Does it? As far as I can tell, it only validates that a characteristicId is set, and that the characteristicType is reasonable (which it has to be anyway since the filter uses the characterisicType to find the value in the first place.
Realistically this is only checking that CharacteristicId != nil (assuming our filter is functioning correctly).
Neither value nor unit are checked, therefore it's a bit of a stretch to say this ensures values are positive and within reasonable ranges.

usecases/eg/lpc/public.go
```diff
@@ -60,6 +60,13 @@ func (e *LPC) ConsumptionLimit(entity spineapi.EntityRemoteInterface) (
 		return
 	}
 
+	// Validate LoadControlLimitData using EG LPC validator
+	// This ensures data integrity and compliance with EG LPC specification requirements
+	if err := EGLPCLoadControlLimitValidator.Validate(value); err != nil {
+		resultErr = api.ErrDataInvalid
+		return
+	}
+
```
sthelen-enqs: This does nothing that isn't already done by lines 58-61. It's not technically wrong, it's just dead code that's burning cycles.

usecases/eg/lpc/public.go
```diff
@@ -123,6 +130,12 @@ func (e *LPC) FailsafeConsumptionActivePowerLimit(entity spineapi.EntityRemoteIn
 		return 0, api.ErrDataNotAvailable
 	}
 
+	// Validate DeviceConfigurationKeyValueData using EG LPC validator
+	// This ensures failsafe power limit values are within reasonable ranges and properly formatted
+	if err := EGLPCDeviceConfigurationValidator.Validate(data); err != nil {
+		return 0, api.ErrDataInvalid
+	}
+
```
sthelen-enqs: Same as above, except that in addition the comment is very misleading. This code doesn't actually check that power limit values are within reasonable limits. It only checks if KeyId is set (which must be set anyway if we've reached this function cal), and if a value is present that only ScaledNumber and/or Duration are set on it (but not String/Boolean/DateTime).

I'm not sure of the utility of this, and from a validation standpoint it's not even technically correct since:
a) data.Value _must_ be set at this point (the code currently passes when it isn't),
b) I don't think the SPINE/EEBUS spec forbid having additional members of the DeviceConfigurationKeyValueValueType set at this point (such as DateTime, Boolean or String), and
c) if the spec did forbit it, the check would also need to be that  ScaledNumber XOR Duration should be set on it since a failsafeConsumptionActivePowerLimit should not include a duration, and a failsafeDurationMinimum should not contain a scaledNumber

usecases/eg/lpc/public.go
sthelen-enqs: Is the intention here to replace the current piecemeal validators with generic validators? In that case this should replace L59-61 and then I'd need to double-check the correctness again.

usecases/internal/measurement.go
```diff
@@ -28,11 +30,19 @@ func MeasurementPhaseSpecificDataForFilter(
 		return nil, api.ErrDataNotAvailable
 	}
 
+	// Validate validator parameter
+	if validator == nil {
+		return nil, fmt.Errorf("validator is required")
+	}
+
```
sthelen-enqs: Could we make validator pass-by-value instead of pointer typed? Then it cannot be nil by type guarantee which reduces the risk of errors at runtime and (imo) makes it easier to reason about.

I don't think the struct is large enough that we have to worry about passing it by value, and I think the go compiler should just do the right thing™

Not a blocker in any way, just wondering if we can improve the ergonomics here.

usecases/internal/measurement.go
```diff
@@ -28,11 +30,19 @@ func MeasurementPhaseSpecificDataForFilter(
...
 	var result []float64
 
 	for _, item := range data {
-		if item.Value == nil || item.MeasurementId == nil {
-			continue
+		// Use validator instead of basic nil checks
+		if err := validator.Validate(&item); err != nil {
+			// For MGCP-003 compliance, we explicitly handle ErrSkipMeasurement
+			// All other validation errors also result in skipping the measurement
+			continue // Skip invalid measurements, don't fail entire operation
```
sthelen-enqs: Although the reasoning is correct, this error message is confusing as no error is handled explicitly.
It might be nice to have some way of logging/warning/reporting measurements with valueState != normal and probably another one for other validationErrors, but I don't know if we need that in this PR.
```suggestion
			// Due to MGCP-003, and MPC-003 we want to skip measurements with valueState != normal
			// We also want to skip all other validation errors so that we don't fail the entire operation
			// just because the remote contains invalid data
			continue
```

usecases/internal/validation.go
```diff
@@ -0,0 +1,350 @@
...
+// Returns the first item that passes all validation rules.
+//
+// Example:
+//
+//	measurements := []model.MeasurementDataType{...}
+//	valid, err := validator.ValidateFirst(measurements)
+//	if err != nil {
+//		return api.ErrDataNotAvailable
+//	}
+func (v *BaseValidator[T]) ValidateFirst(items []T) (T, error) {
```
sthelen-enqs: I dislike the name of this function. The function is not validating the first element of items[], it's finding the first valid element.
```suggestion
func (v *BaseValidator[T]) FindFirstValidItem(items []T) (T, error) {
```

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+//	if err != nil {
+//		return 0, api.ErrDataNotAvailable
+//	}
+//	return value, nil
+func GetMeasurementValue(measurements []model.MeasurementDataType, validator *MeasurementValidator) (float64, error) {
+	// Convert slice to pointer slice for validation
+	ptrMeasurements := make([]*model.MeasurementDataType, len(measurements))
+	for i := range measurements {
+		ptrMeasurements[i] = &measurements[i]
+	}
```
sthelen-enqs: I don't see why the conversion here is necessary, but I'm not that bothered by it either.

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+		"Value",
+	)
+}
+
+// RequireValueType ensures ValueType matches expected type
+func RequireValueType(expected model.MeasurementValueTypeType) ValidationRule[*model.MeasurementDataType] {
+	return func(m *model.MeasurementDataType) error {
+		if expected == "" {
+			return nil // Skip if no expected type specified
+		}
```
sthelen-enqs: This feels like a caller error that we shouldn't be silently ignoring here, if somebody calls this function incorrectly and it isn't doing what they're expecting they should get an error

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+		return nil
+	}
+}
+
+// RequireValueSource ensures ValueSource is one of allowed types
+func RequireValueSource(allowed ...model.MeasurementValueSourceType) ValidationRule[*model.MeasurementDataType] {
+	if len(allowed) == 0 {
+		return func(m *model.MeasurementDataType) error { return nil }
+	}
```
sthelen-enqs, line 199: This feels like a caller error that we shouldn't be silently ignoring here.

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+		if len(allowed) == 0 {
+			return nil // Any value is acceptable if no restrictions
+		}
+		
+		// Validate against allowed values
+		for _, allowed := range allowed {
+			if *m.ValueSource == allowed {
+				return nil
+			}
+		}
```
sthelen-enqs:
```suggestion
		// Validate against allowed values
		if slices.Contains(allowed, *m.ValueSource) {
			return nil
		}
```

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+var MGCPPowerValidator = NewMeasurementValidator().
+	WithName("MGCP Power (Scenario 2)").
+	WithRule(RequireMeasurementId()).                              // M: Mandatory
+	WithRule(RequireMeasurementValue()).                           // M: Mandatory
+	WithRule(RequireValueType(model.MeasurementValueTypeTypeValue)). // M: Mandatory = "value"
+	WithRule(RequireValueSource(                                   // R: Recommended
+		model.MeasurementValueSourceTypeMeasuredValue,
+		model.MeasurementValueSourceTypeCalculatedValue,
+		model.MeasurementValueSourceTypeEmpiricalValue,
+	)).
```
sthelen-enqs: Afaict valueSource is mandatory for the MU, as this is validator only used in the MA we can (should?) validate that valueSource is set _and_ set correctly.
The same applies for valueSource in the following data points as well.

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+	WithRule(ValidateMeasurementRange(0, 1000)) // 0-1000V reasonable range
+
+// FrequencyMeasurementValidator validates frequency measurements
+//
+// DEPRECATED: Use MGCPFrequencyValidator for MGCP implementations.
+var FrequencyMeasurementValidator = NewMeasurementValidator().
+	WithName("Frequency Measurement (Legacy)").
+	WithRule(RequireMeasurementId()).
+	WithRule(RequireMeasurementValue()).
+	WithRule(RequireValueType(model.MeasurementValueTypeTypeValue))
```
sthelen-enqs: I don't see the point in having these here. They're not being used in the stack, if anyone needs them for some reason they can (and should?) add them in their code.
The Voltage one also somehow gained a 0-1000V range which I'm pretty sure wasn't in the old public.go code

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+	WithRule(RequireMeasurementId()).                              // M: Mandatory
+	WithRule(RequireMeasurementValue()).                           // M: Mandatory
+	WithRule(RequireValueType(model.MeasurementValueTypeTypeValue)). // M: Mandatory = "value"
+	WithRule(RequireValueSource(                                   // R: Recommended
+		model.MeasurementValueSourceTypeMeasuredValue,
+		model.MeasurementValueSourceTypeCalculatedValue,
+		model.MeasurementValueSourceTypeEmpiricalValue,
+	)).
+	WithRule(SkipValueState()) // R: Recommended with MGCP-003
+
```
sthelen-enqs: I just noticed that all of these exist in ma/mgcp/validators.go as well. These ones are duplicate and unused?
If so they should be removed.

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+// RequireValueSourceMandatory ensures ValueSource is present and one of allowed types
+func RequireValueSourceMandatory(allowed ...model.MeasurementValueSourceType) ValidationRule[*model.MeasurementDataType] {
+	return func(m *model.MeasurementDataType) error {
+		if m.ValueSource == nil {
+			return fmt.Errorf("ValueSource is required")
+		}
+		
+		if len(allowed) == 0 {
+			return nil // Any value is acceptable if no restrictions
+		}
```
sthelen-enqs: I don't like the check for empty allowed here. If you need a field to exist and not be nil you can use RequireField, if you're using RequireValieSourceMandatory my expectation is that you have some restrictions on your valueSource and if you forget them it should probably be an error.

usecases/internal/measurement.go
```diff
@@ -62,16 +72,401 @@ func MeasurementPhaseSpecificDataForFilter(
...
+		// Validate against allowed values
+		for _, allowed := range allowed {
+			if *m.ValueSource == allowed {
+				return nil
+			}
+		}
+		
+		return fmt.Errorf("ValueSource must be one of %v, got %s", allowed, *m.ValueSource)
+	}
+}
```
sthelen-enqs: This function as a whole is confusing. This is just a RequireField().ValidateEnum() chained call that unrolls the chained function call.
I don't mind the unrolling, it's just somewhat arbitrary which functions have an unrolled variant and which don't.
e.g. ValidateRange does not have a ValidateRangeMandatory unrolled variant and instead expects you to call RequireScaledNumber().ValidateRange() manually.

usecases/ma/mgcp/validators.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// RequireValueSourceMGCP validates ValueSource when mandatory per MGCP spec
+func RequireValueSourceMGCP(mandatory bool) internal.ValidationRule[*model.MeasurementDataType] {
+	if mandatory {
+		return internal.RequireValueSourceMandatory(mgcpValueSources...)
+	}
+	// For recommended, we validate if present but don't require
+	return internal.RequireValueSource(mgcpValueSources...)
```
sthelen-enqs: Afaict for the GCP valueSource is always Mandatory for the GCP and this should always be internal.RequireValueSourceMandatory(mgcpValueSources...) or if we drop that function to make it more uniform internal.CombineRules(internal.RequireField(...), internal.RequireValueSource(...))

usecases/ma/mgcp/validators.go
```diff
@@ -0,0 +1,114 @@
...
+// MGCPPowerValidator validates power measurements for Scenario 2
+// - valueType: M (Mandatory) = "value"
+// - valueSource: R (Recommended) = measuredValue|calculatedValue|empiricalValue
+// - valueState: R (Recommended) with MGCP-003 rule
+var MGCPPowerValidator = internal.NewMeasurementValidator().
+	WithName("MGCP Power").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(RequireValueSourceMGCP(false)). // Recommended, not mandatory
```
sthelen-enqs: afaict valueSource is always M for the GCP

usecases/ma/mgcp/validators.go
```diff
@@ -0,0 +1,114 @@
...
+// MGCPCurrentValidator validates current measurements for Scenario 5
+// - valueType: M (Mandatory) = "value"
+// - valueSource: R (Recommended) = measuredValue|calculatedValue|empiricalValue
+// - valueState: R (Recommended) with MGCP-003 rule
+var MGCPCurrentValidator = internal.NewMeasurementValidator().
+	WithName("MGCP Current").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(RequireValueSourceMGCP(false)). // Recommended, not mandatory
```
sthelen-enqs: afaict valueSource is always M for the GCP

usecases/ma/mgcp/validators.go
```diff
@@ -0,0 +1,114 @@
...
+// MGCPVoltageValidator validates voltage measurements for Scenario 6
+// - valueType: M (Mandatory) = "value"
+// - valueSource: R (Recommended) = measuredValue|calculatedValue|empiricalValue
+// - valueState: R (Recommended) with MGCP-003 rule
+var MGCPVoltageValidator = internal.NewMeasurementValidator().
+	WithName("MGCP Voltage").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(RequireValueSourceMGCP(false)). // Recommended, not mandatory
```
sthelen-enqs: afaict valueSource is always M for the GCP

usecases/ma/mgcp/validators.go
```diff
@@ -0,0 +1,114 @@
...
+// MGCPFrequencyValidator validates frequency measurements for Scenario 7
+// - valueType: M (Mandatory) = "value"
+// - valueSource: R (Recommended) = measuredValue|calculatedValue|empiricalValue
+// - valueState: R (Recommended) with MGCP-003 rule
+var MGCPFrequencyValidator = internal.NewMeasurementValidator().
+	WithName("MGCP Frequency").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(RequireValueSourceMGCP(false)). // Recommended, not mandatory
```
sthelen-enqs: afaict valueSource is always M for the GCP

usecases/ma/mgcp/validators.go
```diff
@@ -0,0 +1,114 @@
...
+	WithRule(SkipValueState())               // MGCP-003 rule
+
+// Legacy validators for backward compatibility (deprecated for MGCP use)
+
+// basicMGCPValidator provides minimal validation for backward compatibility
+// Deprecated: Use scenario-specific validators (MGCPPowerValidator, etc.) for proper MGCP compliance
+var basicMGCPValidator = internal.NewMeasurementValidator().
+	WithName("MGCP Basic").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue())
```
sthelen-enqs:
```suggestion
```

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	voltageSourceTypes = []model.MeasurementValueSourceType{
+		model.MeasurementValueSourceTypeMeasuredValue,
+		model.MeasurementValueSourceTypeCalculatedValue,
+	}
+
+	// frequencySourceTypes are the allowed value sources for frequency measurements
+	frequencySourceTypes = []model.MeasurementValueSourceType{
+		model.MeasurementValueSourceTypeMeasuredValue,
+		model.MeasurementValueSourceTypeCalculatedValue,
+	}
```
sthelen-enqs: Does it mention anywhere in the spec that empiricalValue isn't valid here?

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// powerValidator validates power measurements
+var powerValidator = internal.NewMeasurementValidator().
+	WithName("MPC Power").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(powerSourceTypes...)).
```
sthelen-enqs: Afaict valueSource is always required for the MU and can be made mandatory here

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false))
+
+// energyValidator validates energy measurements (consumption and production)
+var energyValidator = internal.NewMeasurementValidator().
+	WithName("MPC Energy").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(energySourceTypes...)).
```
sthelen-enqs: Afaict valueSource is always required for the MU and can be made mandatory here

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false))
+
+// currentValidator validates current measurements
+var currentValidator = internal.NewMeasurementValidator().
+	WithName("MPC Current").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(currentSourceTypes...)).
+	WithRule(internal.ValidateValueState("", true)) // Any state required per spec
```
sthelen-enqs: Afaict valueSource is always required for the MU and can be made mandatory here
I, however, see nothing in the MU that makes valueState mandatory for current measurements, I also don't know why there's an exception here allowing "any state".
Because of MPC-003 shouldn't this be an equivalent SkipValueState as in mgcp/validators?

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+// powerValidator validates power measurements
+var powerValidator = internal.NewMeasurementValidator().
+	WithName("MPC Power").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(powerSourceTypes...)).
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false))
```
sthelen-enqs: Because of MPC-003 shouldn't this be an equivalent SkipValueState as in mgcp/validators?

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+
+// energyValidator validates energy measurements (consumption and production)
+var energyValidator = internal.NewMeasurementValidator().
+	WithName("MPC Energy").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(energySourceTypes...)).
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false))
```
sthelen-enqs: Because of MPC-003 shouldn't this be an equivalent SkipValueState as in mgcp/validators?

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.ValidateValueState("", true)) // Any state required per spec
+
+// voltageValidator validates voltage measurements
+var voltageValidator = internal.NewMeasurementValidator().
+	WithName("MPC Voltage").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(voltageSourceTypes...)).
+	WithRule(internal.ValidateMeasurementRange(0, 1000)) // 0-1000V per spec
```
sthelen-enqs: I don't see anything in the spec limiting us to 0-1000V and see no reason to do validate this here.

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+
+// frequencyValidator validates frequency measurements
+var frequencyValidator = internal.NewMeasurementValidator().
+	WithName("MPC Frequency").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(frequencySourceTypes...)).
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false)) // Reject error states
```
sthelen-enqs: Because of MPC-003 shouldn't this be an equivalent SkipValueState as in mgcp/validators?

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.RequireValueSource(energySourceTypes...)).
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false))
+
+// currentValidator validates current measurements
+var currentValidator = internal.NewMeasurementValidator().
+	WithName("MPC Current").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(currentSourceTypes...)).
```
sthelen-enqs: Afaict valueSource is always required for the MU and can be made mandatory here

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.ValidateValueState("", true)) // Any state required per spec
+
+// voltageValidator validates voltage measurements
+var voltageValidator = internal.NewMeasurementValidator().
+	WithName("MPC Voltage").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
```
sthelen-enqs: Afaict valueSource is always required for the MU and can be made mandatory here

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.RequireValueSource(voltageSourceTypes...)).
+	WithRule(internal.ValidateMeasurementRange(0, 1000)) // 0-1000V per spec
+
+// frequencyValidator validates frequency measurements
+var frequencyValidator = internal.NewMeasurementValidator().
+	WithName("MPC Frequency").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(frequencySourceTypes...)).
```
sthelen-enqs: Afaict valueSource is always required for the MU and can be made mandatory here

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.ValidateValueState(model.MeasurementValueStateTypeNormal, false)) // Reject error states
+
+// getMeasurementValue is a helper that validates and extracts the value from measurements
+// DEPRECATED: Use MeasurementPhaseSpecificDataForFilter with validators instead
+func getMeasurementValue(measurements []model.MeasurementDataType, validator *internal.MeasurementValidator) (float64, error) {
+	return internal.GetMeasurementValue(measurements, validator)
+}
```
sthelen-enqs: If this is deprecated, can we just get rid of it?

usecases/ma/mpc/validators.go
```diff
@@ -0,0 +1,92 @@
...
+	WithRule(internal.RequireValueSource(currentSourceTypes...)).
+	WithRule(internal.ValidateValueState("", true)) // Any state required per spec
+
+// voltageValidator validates voltage measurements
+var voltageValidator = internal.NewMeasurementValidator().
+	WithName("MPC Voltage").
+	WithRule(internal.RequireMeasurementId()).
+	WithRule(internal.RequireMeasurementValue()).
+	WithRule(internal.RequireValueType(model.MeasurementValueTypeTypeValue)).
+	WithRule(internal.RequireValueSource(voltageSourceTypes...)).
```
sthelen-enqs: Because of MPC-003 shouldn't this be an equivalent SkipValueState as in mgcp/validators?

sthelen-enqs (changes requested):
I have to admit I'm not in love with these changes. The validator logic feels more complicated than necessary, and I fear will make it easier to filter data incorrectly as evidenced by the nits and spec inconsistencies I've found so far.
I, however, do not have a better idea as of right now either.

usecases/eg/lpc/public.go
```diff
@@ -298,16 +285,10 @@ func (e *LPC) ConsumptionNominalMax(entity spineapi.EntityRemoteInterface) (floa
 	data, err := electricalConnection.GetCharacteristicsForFilter(filter)
 	if err != nil {
 		return 0, err
-	} else if len(data) == 0 || data[0].Value == nil {
+	} else if len(data) == 0 || data[0].CharacteristicId == nil || data[0].Value == nil {
```
andig: Out of curiosity: shouldn‘t these checks also guard against length != 1?

PR #183, opened 2025-10-13
Rename DataUpdateCurrentsPerPhase to DataUpdateCurrentPerPhase (BC)

### LINKED ISSUE #182: Minor inconsistency between MPC and MGCP event names
Consider this difference (current vs currents):

```
mpc.DataUpdateCurrentsPerPhase
mgcp.DataUpdateCurrentPerPhase
```

vs

```
mpc.DataUpdateVoltagePerPhase
mgcp.DataUpdateVoltagePerPhase
```

Not sure if this is from the spec or internal. If you want to let me know which variant you prefer (`DataUpdateCurrentPerPhase` in line with voltages) I'd be happy to open PR and also check remaining use cases.

Note that the interface is also `mpc.CurrentPerPhase`. The event name seems to be the odd outlier.

andig:
Fix https://github.com/enbility/eebus-go/issues/182

### DIFF
--- a/usecases/ma/mpc/events.go
+++ b/usecases/ma/mpc/events.go
@@ -113,7 +113,7 @@ func (e *MPC) deviceMeasurementDataUpdate(payload spineapi.EventPayload) {
 		// Scenario 3
 		filter.ScopeType = util.Ptr(model.ScopeTypeTypeACCurrent)
 		if measurement.CheckEventPayloadDataForFilter(payload.Data, filter) && e.EventCB != nil {
-			e.EventCB(payload.Ski, payload.Device, payload.Entity, DataUpdateCurrentsPerPhase)
+			e.EventCB(payload.Ski, payload.Device, payload.Entity, DataUpdateCurrentPerPhase)
 		}
 
 		// Scenario 4
--- a/usecases/ma/mpc/types.go
+++ b/usecases/ma/mpc/types.go
@@ -41,7 +41,7 @@ const (
 	// Use `CurrentPerPhase` to get the current data
 	//
 	// Use Case MCP, Scenario 3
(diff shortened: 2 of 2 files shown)


andig:
@sthelen-enqs this is a breaking chore fix that would help improve naming consistency

uhl:
@andig Thanks for the contribution— you're right that CurrentPerPhase is the correct name and Currents is the outlier (the MPC interface method is already CurrentPerPhase).
I'd hold off on merging this as a hard rename though. DataUpdateCurrentsPerPhase is part of the public API and I can see at least a handful of downstream consumers on GitHub. A cosmetic rename that forces every one of them to touch their import sites on the next release feels like a poor tradeoff for the ergonomic win — especially since the CONTRIBUTING guideline explicitly discourages purely cosmetic patches, and there's no changelog convention here to loudly signal the break.
What do you think?

andig:
@uhl thank you for the comment. I agree this is a breaking change. However- with 0.x versions breaking changes are to be expected. The issue here is that it makes the library usage confusing and imho more than a cosmetic rename. PR (and commit) is already named `BC` to signal that.

sthelen-enqs:
After some more discussion, we've decided not to go ahead with this here. The rename is a relatively minor change and the name isn't so wrong that we think the break is worth it right now.
Thanks for the PR though

andig:
I take the decision obviously but it's still unfortunate. I personally consider a library being non-surprising to consumers as a high value. The current branch is explicitly called `dev` and the version number is coming up to 0.8. That is exactly the best place where to make these changes. And obviously, breaking changes are absolutely expected as https://github.com/enbility/eebus-go/pull/210 shows.

Issue #116, opened 2024-10-04
LPC and LPP dependencies to MPC or MGCP


DerAndereAndi:
An LPC/LPP server component (actor "Controllable System") needs to provide a measurementId to a limit. If those are not supported, an unused measurement ID should be used and hope that the opposite site doesn't check for validity of the provided ID.

To automate this, the LPC/LPP usecases should be initialized AFTER MPC and MGCP by the application. Then the use case implementation can check for such an ID automatically and use it.

So either the application has the responsibility to do this right, or we change the way usecases are added. An option would be to add all usecases in a single step instead of individually, and there this kind of logic can be implemented.

sthelen-enqs:
Fixed by #186

PR #186, opened 2025-10-29
Add LPC/LPP dependency check on MPC/MGCP use cases

mohamedeltawel:
LPC/LPP use-cases now detect MPC/MGCP support and, when available, use their predefined measurementId, ElectricalConnectionId, and ParameterId; otherwise they fall back to default IDs.
This fixes issue #116


DerAndereAndi (approved):

DerAndereAndi:
Thanks @mohamedeltawel this looks good!

PR #189, opened 2025-11-26
[lpc/lpp eg] subscription and read request for scenario 4

kirollosnct:
This PR adds missing pre-scenario and initial scenario communication for LPC/LPP EnergyGuard scenario 4, which requires to subscribe on ElectricalConnection feature and trigger a read request for electricalConnectionCharacteristicListData.


sthelen-enqs (approved):
LGTM, thanks

DerAndereAndi:
Shouldn't we also check if scenario 4 is even announced as part of the usecase using `IsScenarioAvailableAtEntity(entity, 4)` ?

sthelen-enqs:
We should look into #198 and how that influences this PR

PR #191, opened 2025-12-12
Use the entityType to determine the characteristicType for eg/lpc and eg/lpp

sthelen-enqs:
Previously, the LPC/LPP code was using the deviceType to determine whether we should check the Power*NominalMax or Contractual*NominalMax data points. It should be more correct to use the EntityType instead since the specification mentions "CEM" explicitly.

This PR also sets the DeviceType in the LPC/LPP Test Suites as this defaulted to nil (and caused issue with the old variant that was checking the DeviceType).


DerAndereAndi (approved):

DerAndereAndi:
Thanks @sthelen-enqs this looks good :)

PR #210, opened 2026-03-16
Add msgCounter to function signature of result callbacks

sthelen-enqs:
This simplifies usage of the result callback as you can use the msgCounter in the callback to determine which message the response correlates to and don't need to provide that information out-of-band.

### DIFF
--- a/examples/ced/main.go
+++ b/examples/ced/main.go
@@ -162,7 +162,7 @@ func (h *controlbox) sendLimit(entity spineapi.EntityRemoteInterface) {
 			Value:    7000,
 		}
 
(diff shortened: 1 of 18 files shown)


DerAndereAndi (changes requested):
Some findings:

- the generated mock files need regeneration
- usecases/internal/loadcontrol_test.go:850 and :1129 update the signature but the callback body doesn't assert that the received msgCounter equals the value returned from Write*.

andig:
@sthelen-enqs this is currently inconsistent as it is not applied to all use cases (e.g. OHPCF)- would you care for follow-up PR?

PR #211, opened 2026-03-16
Implement the MGCP use case for the GCP

sthelen-enqs:
This adds an implementation of the MGCP use case for the Grid Connection Point. This PR builds on top of #108.

Draft until #108 is merged


sthelen-enqs:
Included in https://github.com/enbility/eebus-go/pull/224

PR #214, opened 2026-04-14
Revert LPP Failsafe values back to positive values in the examples

uhl:
Per the spec, the values are always >= 0 because their name already indicates the production direction Only the APPL value itself (the LoadLimit.Value in SetProductionLimit and WriteProductionLimit) remains <= 0 (-10000 in HEMS, -100 in controlbox), since that's the SignDependentAbsValueLimit where sign matters.


sthelen-enqs:
LGTM, thanks

PR #218, opened 2026-05-12
Fix wrong parameter description selection in LoadControlLimits

### LINKED ISSUE #217: WriteLoadControlPhaseLimits may pick the wrong parameter description when a device exposes multiple measurement types per phase
usecases/internal/loadcontrol.go
 Function: WriteLoadControlPhaseLimits (line 222), also affects LoadControlLimits (line 22)

  ### Summary

  The filter used to look up the phase's ElectricalConnectionParameterDescription constrains only on phase name, not on measurement type. If a remote device exposes more than one
  description per phase (e.g. current and voltage), the code silently uses elParamDesc[0] and may resolve the limit against the wrong MeasurementId/ParameterId.

  ### Details

  At line 247:
```

  filter2 := model.ElectricalConnectionParameterDescriptionDataType{
      AcMeasuredPhases: util.Ptr(phaseLimit.Phase),
  }
```

 OPEV/OSCEV limits are always in amperes, so the intended entry is the current description. If [0] is a voltage or power description, the write targets the wrong LimitId or gets clamped against the wrong permitted range.

Not observed on tested hardware (Elli, Porsche MCC, SMA HEM 2.0) — they only expose current descriptions in this list — but a conformant device is allowed to expose more.
LoadControlLimits (line 50) has the same filter and inherits the issue.

Sidenote:
 - Missing nil check on ParameterId. Line 248 guards MeasurementId != nil, but line 280 dereferences *elParamDesc[0].ParameterId unchecked. ParameterId is optional; a remote omitting it would panic. Same gap in LoadControlLimits at line 84.

  ### Reproduction
Extend the fixture at loadcontrol_test.go:915 so one phase has two parameter descriptions (current and voltage), with voltage listed first. The resolved LimitId will be wrong.

andig:
Fixes #217.

Asked Claude for a fix since [evcc-io/evcc](https://github.com/evcc-io/evcc) uses both opev and oscev. Fix is LGTM.

## Problem

`LoadControlLimits` and `WriteLoadControlPhaseLimits` in `usecases/internal/loadcontrol.go` looked up `ElectricalConnectionParameterDescription` entries filtered only by `AcMeasuredPhases` and then blindly used `elParamDesc[0]`. If a remote device exposes more than one description per phase (e.g. current and voltage), the wrong entry could be picked and the limit would be resolved against an unrelated `MeasurementId` / `ParameterId`. Additionally, `*elParamDesc[0].ParameterId` was dereferenced without a nil guard on the write side, which could panic against a conformant remote that omits the optional `ParameterId`.

Not observed on Elli, Porsche MCC, or SMA HEM 2.0 (they only expose current descriptions), but a conformant device is allowed to expose more — and the bug surfaces whenever a non-current description is listed first.

## Fix

Walk all parameter descriptions for the phase and pick the one whose `MeasurementId` is referenced by one of the limit descriptions, then use that description's `ParameterId` for the permitted-value lookup / adjustment. Adds the missing `ParameterId` nil guard on the write path.

The four public entry points are all thin wrappers over the two internal helpers and inherit the fix:

- `opev.OPEV.LoadControlLimits` / `opev.OPEV.WriteLoadControlLimits`
- `oscev.OSCEV.LoadControlLimits` / `oscev.OSCEV.WriteLoadControlLimits`

### DIFF
--- a/usecases/internal/loadcontrol.go
+++ b/usecases/internal/loadcontrol.go
@@ -50,25 +50,39 @@ func LoadControlLimits(
 		filter := model.ElectricalConnectionParameterDescriptionDataType{
 			AcMeasuredPhases: &phaseName,
 		}
-		elParamDesc, err := evElectricalConnection.GetParameterDescriptionsForFilter(filter)
-		if err != nil || len(elParamDesc) == 0 || elParamDesc[0].MeasurementId == nil {
+		elParamDescs, err := evElectricalConnection.GetParameterDescriptionsForFilter(filter)
+		if err != nil || len(elParamDescs) == 0 {
 			// there is no data for this phase, the phase may not exist
 			// so do not add id to the result
 			continue
 		}
 
+		// A phase may have several parameter descriptions (e.g. current, voltage, power).
+		// Pick the one whose MeasurementId is referenced by one of the limit descriptions —
+		// blindly using elParamDescs[0] would resolve the limit against the wrong measurement
+		// type when a conformant remote lists them in an unexpected order.
+		var paramDesc *model.ElectricalConnectionParameterDescriptionDataType
 		var limitDesc *model.LoadControlLimitDescriptionDataType
-		for _, desc := range limitDescriptions {
-			if desc.MeasurementId != nil &&
-				elParamDesc[0].MeasurementId != nil &&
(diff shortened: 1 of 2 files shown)


sthelen-enqs (approved):
LGTM, thanks

PR #223, opened 2026-06-08
Implement cem/ohpcf

sthelen-enqs:


andig:
OT: I'm wondering how useful OHPCF, especially compared to FLOA, really is in real-live scenarios. How does it help if the heatpump announces 5kW for unknown runtime if your PV only delivers 4kW?

usecases/api/cem_ohpcf.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Schedule an optional power consumption process [OHPCF-004].
+	//
+	// note:
+	// Rescheduling an already scheduled power consumption process is possible as long as the
+	// scheduled process has not startet yet.
+	//
+	// parameters:
+	//   - start: The start time of the power consumption
+	SchedulePowerConsumptionProcess(entity spineapi.EntityRemoteInterface, start time.Time, resultCB func(result model.ResultDataType)) (*model.MsgCounterType, error)
```
andig, line 75: This hard-codes `SchedulePowerConsumptionProcess` to using `model.NewAbsoluteOrRelativeTimeTypeFromTime(start)` which doesn't work on Vaillant which only accepts durations. From the AbsoluteOrRelativeTimeType — EEBus_SPINE_TS_CommonDataTypes.xsd, lines 51–53:

```xml
<xs:simpleType name="AbsoluteOrRelativeTimeType">
    <xs:union memberTypes="xs:duration xs:dateTime"/>
</xs:simpleType>
```

I'd be happy to provide a PR. Options:

- have both time and duration variants of the function
- make start a variadic option (`WithTime` vs `WithDuration`)
- simplify have the actual model type passed in (via interface or any)

Which one would you prefer?

andig: It seems passing the start time as `*model.AbsoluteOrRelativeTimeType` is the absolute easiest way of doing this.

andig: Took another look:

```
┌──────────────────────────────────────┬─────────────────────────────────┬───────────────────────────────┬────────────────────────────────────────────────────────┐
│               Use case               │              Field              │        Encoding chosen        │                      constructor                       │
├──────────────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────┤
│ cs/lpc/public.go:81                  │ LoadControl limit EndTime       │ relative                      │ ...FromDuration(limit.Duration)                        │
├──────────────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────┤
│ cs/lpp/public.go:83                  │ LoadControl limit EndTime       │ relative                      │ ...FromDuration(limit.Duration)                        │
├──────────────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────┤
│ usecases/internal/loadcontrol.go:162 │ LoadControl limit EndTime       │ relative                      │ ...FromDuration(limit.Duration)                        │
├──────────────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────┤
│ cem/cevc/public_scen2.go:111-130     │ TimeSeries Start/End            │ relative (incl. literal       │ ...FromDuration /                                      │
│                                      │                                 │ "PT0S")                       │ NewAbsoluteOrRelativeTimeType("PT0S")                  │
├──────────────────────────────────────┼─────────────────────────────────┼───────────────────────────────┼────────────────────────────────────────────────────────┤
│ cem/ohpcf/public.go:308              │ PowerSequence Schedule          │ absolute                      │ ...FromTime(start) ← outlier                           │
│                                      │ StartTime                       │                               │                                                        │
└──────────────────────────────────────┴─────────────────────────────────┴───────────────────────────────┴────────────────────────────────────────────────────────┘
```

Rest of the library is using duration. This is the _only_ place where actual time seems to be used. So the idiomatic fix- without adding additional flexibility would be changing `start` into `startIn` as `time.Duration`.

wdyt?

andig: That should settle it:

> 3.1.8.2 Rules regarding the usage of time-related information
All time-related information within this Use Case SHALL be presented as relative times. This means
absolute times SHALL NOT be used. All start- and end-times SHALL be interpreted as relative to the
time the message was transmitted.

andig:
> OT: I'm wondering how useful OHPCF, especially compared to FLOA, really is in real-live scenarios. How does it help if the heatpump announces 5kW for unknown runtime if your PV only delivers 4kW?

The FLOA spec hasn‘t even been released, much less implemented…

andig:
@sthelen-enqs fwiw I can confirm this works against Vaillant heatpump with the proposed changes. It would be great to get both merged! If you need anything specifically tested pls let me know.

sungho1shin:
Thanks for this! I tried to interoperate this CEM implementation against an
independent OHPCF **Compressor** (a `SmartEnergyManagementPs` *server* that
correctly advertises the OHPCF use case — `optimizationOfSelfConsumptionByHeat
PumpCompressorFlexibility` v1.0.0, `scenarioSupport [1,2]`, actor `Compressor`,
and serves `smartEnergyManagementPsData` with read+write).

I hit two issues on the current branch head (`abdcf86`). The first is a hard
blocker; the second is a spec-conformance bug. A `ma/mpc` client running in the
same binary against the same peer worked perfectly, which isolates both issues to
`cem/ohpcf`.

## 1. `cem/ohpcf`'s `HandleEvent` is never invoked → no subscription/read ever happens (blocker)

Symptoms against the server:
- Every `OptionalPowerConsumption` / `PowerConsumptionProcessState` / … call returns `ErrDataNotAvailable`.
- Every Scenario 2 call (`Schedule/Pause/Resume/Abort…`) fails with `data not available` (they call `OptionalPowerConsumption` first).
- On the wire the CEM sends subscriptions only for `NodeManagement`, `ElectricalConnection`, `Measurement` — **no `SmartEnergyManagementPs` subscription or read is ever sent**, and the server receives zero `smartEnergyManagementPsData` requests.

Root cause: `NewOHPCF` never subscribes the concrete use case to SPINE events, so
`(*OHPCF).HandleEvent` is never called (I confirmed with an `fmt.Println` at the
top of `HandleEvent` — it never fires, while `ma/mpc`'s does). `NewUseCaseBase`
only subscribes the embedded base, and Go has no virtual dispatch through
embedding, so the base's `HandleEvent` runs instead of the use case's. Every
other use case adds the missing line — e.g. `usecases/ma/mpc/usecase.go`:

```go
uc := &MPC{ UseCaseBase: usecase }
_ = localEntity.Device().Events().Subscribe(uc)   // <-- present in ma/mpc
return uc
```

Note this is entirely local to the CEM and independent of the remote device, so
`cem/ohpcf` as published cannot drive its own subscribe/bind/read path against
*any* peer.

**Fix** — `usecases/cem/ohpcf/usecase.go`:

```diff
 	uc := &OHPCF{
 		UseCaseBase: usecase,
 	}
 
+	// register the concrete use case for SPINE events so its own HandleEvent
+	// runs (as in every other use case, e.g. usecases/ma/mpc/usecase.go)
+	_ = localEntity.Device().Events().Subscribe(uc)
+
 	return uc
 }
```

I'd also suggest requesting the current data in `connected()` so the cache is
populated when the process already exists at connect time (otherwise it's only
filled on the next server-side change) — matching the `ma/mpc` pattern. In
`usecases/cem/ohpcf/events.go`:

```diff
 		if !semp.HasBinding() {
 			if _, err := semp.Bind(); err != nil {
 				logging.Log().Debug(err)
 			}
 		}
+
+		if _, err := semp.RequestData(); err != nil {
+			logging.Log().Debug(err)
+		}
 	}
 }
```

With both changes, Scenario 1 (`available`/`running`/`scheduled`/`completed`/
`stopped`) and Scenario 2 (`schedule`/`resume`/`stop`) work end-to-end.

## 2. `paused` state mapping doesn't match the spec

After fixing #1, the only remaining failure is the paused state:
`PowerConsumptionProcessState` returns `ErrNotSupported` when the process is
paused.

`usecases/cem/ohpcf/public.go` maps paused from `PowerSequenceStateTypeScheduled
Paused` (`"scheduledPaused"`), but [OHPCF-012/2/3] and the mapping table specify
the string `"paused"` (`PowerSequenceStateTypePaused`) — which is also exactly
what `PausePowerConsumptionProcess` *writes*. So the current code writes
`"paused"` but reads back expecting `"scheduledPaused"`, which is internally
inconsistent and returns `ErrNotSupported` for a paused process.

**Fix** — `usecases/cem/ohpcf/public.go`:

```diff
 		case model.PowerSequenceStateTypeRunning:
 			return ucapi.CompressorPowerConsumptionStateRunning, nil
-		case model.PowerSequenceStateTypeScheduledPaused:
+		case model.PowerSequenceStateTypePaused:
 			return ucapi.CompressorPowerConsumptionStatePaused, nil
 		case model.PowerSequenceStateTypeCompleted:
```

(If you also want to accept `scheduledPaused`, add it as an extra case rather
than replacing `paused`.)

## How to reproduce

1. Build a CEM with `cem/ohpcf` + `ma/mpc` and pair it with any device exposing a
   `SmartEnergyManagementPs` **server** on a `Compressor` entity that advertises
   OHPCF scenarios 1 & 2.
2. Add `fmt.Println("HandleEvent", payload.EventType)` at the top of
   `(*OHPCF).HandleEvent` → observe it never prints (while `ma/mpc`'s handler
   does).
3. Call any `OptionalPowerConsumption`/`PowerConsumptionProcessState` → always
   `ErrDataNotAvailable`; a SPINE trace shows no `smartEnergyManagementPsData`
   subscription/read.
4. Apply fix #1 → data flows; drive the compressor through its states and observe
   all map correctly except paused, which surfaces #2; apply fix #2 → paused maps
   correctly.

Happy to open a PR with these two changes if useful.

andig:
@sungho1shin did you test against https://github.com/enbility/eebus-go/pull/228?

sungho1shin:
@andig  Good call — no, my earlier report was against the `implement-cem-ohpcf` head
(`abdcf86`) on its own, without #228. I've now retested with **#228 applied on
top**, against the same independent OHPCF `SmartEnergyManagementPs` server.

#228 fixes the blocker I hit: `(*OHPCF).HandleEvent` is now registered
(`Events().Subscribe(uc)`) and `connected()` calls `RequestData()`, so the CEM
subscribes and reads the data. Scenario 1 and Scenario 2 now work end-to-end:

| compressor state | `PowerConsumptionProcessState` |
|---|---|
| inactive  | `available` Pass |
| scheduled | `scheduled` Pass |
| running   | `running` Pass |
| completed | `completed` Pass |
| invalid   | `stopped` Pass |
| **paused**    | **`ErrNotSupported` Fail** |

The one remaining issue (independent of #228) is the `paused` mapping in
`PowerConsumptionProcessState` (`usecases/cem/ohpcf/public.go`): it matches
`PowerSequenceStateTypeScheduledPaused` (`"scheduledPaused"`), but [OHPCF-012/2/3]
specifies the state string `"paused"` — which is exactly what this use case's own
`PausePowerConsumptionProcess` writes. So a paused process is written as
`"paused"` and then read back as unsupported.

```diff
 		case model.PowerSequenceStateTypeRunning:
 			return ucapi.CompressorPowerConsumptionStateRunning, nil
-		case model.PowerSequenceStateTypeScheduledPaused:
+		case model.PowerSequenceStateTypePaused:
 			return ucapi.CompressorPowerConsumptionStatePaused, nil
 		case model.PowerSequenceStateTypeCompleted:
```

(If some devices report `scheduledPaused`, adding it as an extra case alongside
`paused` would cover both.)

With #228 + that one line, all six states map correctly on my side. 
Thanks for the quick turnaround on #228!

andig:
Added fix to #228 @sthelen-enqs

andig:
@sthelen-enqs confirmed working with addition of #228. It would be great if we could get both merged!

sthelen-enqs:
Thanks for the testing, and the fixes

PR #224, opened 2026-06-09
Implement the MPC and MGCP use case for Monitored Units

sthelen-enqs:
This adds an implementation of the MPC use case for Monitored Units.

This PR is based on https://github.com/enbility/eebus-go/pull/108, rebased for current dev.


andig:
@sthelen-enqs this has apparently just broken https://github.com/evcc-io/evcc/issues/31561 since EntityTypeTypeGridGuard from CsLPC/CsLPP validEntityTypes

PR #226, opened 2026-06-25
Add Monitoring of DHW Temperature (MDT) use case

andig:
Adds the `ma/mdt` use case: a Monitoring Appliance reading a DHW circuit's domestic hot water temperature, per the EEBus UC TS *Monitoring of DHW Temperature* v1.0.0 (`monitoringOfDhwTemperature`, actors DHW Circuit ↔ Monitoring Appliance, scenario 1, `Measurement` feature, scope `dhwTemperature`).

Replaces #123 (closed) and implements its review feedback:

- **Value state handling** — non-normal readings (`outOfRange` / `error`) are ignored and reported as `ErrDataInvalid`, as required by the UC spec ([MDT-005]).
- **Unit handling** — `Temperature(entity, unit)` now takes the requested unit (`degC` / `degF` / `K`) and converts from the unit announced in the measurement description, instead of returning a bare value in an unknown unit.

@sthelen-enqs this would be nice with OHPCF.


andig:
@Ka0o0 this PR is the missing piece for usable OHPCF in practice. Review would be highly appreciated.

volschin:
Maybe the rest of the DHW case is interesting for you also.
https://github.com/volschin/eebus-ha-bridge/commit/8ff15b574a72372b4afcf291fab64a6c98f3da64

andig:
> Maybe the rest of the DHW case is interesting for you also. [volschin/eebus-ha-bridge@8ff15b5](https://github.com/volschin/eebus-ha-bridge/commit/8ff15b574a72372b4afcf291fab64a6c98f3da64)

@volschin what exactly do you mean, or rather which UC? Seems you are referring to OHPCF?

volschin:
@andig best to look here in the first section https://wiki.my-gekko.com/de/index/gerate/vaillant/handbuchvail/unterstutzte/unterstutzte.html

andig:
Ah, you mean the EEBus HVAC System Function Package?

usecases/ma/mdt/public.go
```diff
@@ -0,0 +1,80 @@
...
+
+	switch to {
+	case model.UnitOfMeasurementTypedegC:
+		return celsius, nil
+	case model.UnitOfMeasurementTypedegF:
+		return celsius*9/5 + 32, nil
+	case model.UnitOfMeasurementTypeK:
+		return celsius + 273.15, nil
+	default:
+		return 0, api.ErrDataInvalid
```
Ka0o0: This error can only be returned by providing an invalid parameter by the developer. api.ErrDataInvalid is typically only used for cases where we receive invalid data from the remote entity. I think a better error would be `api.ErrNotSupported` here.

Ka0o0 (commented):
Thank you, looks good so far. I could test the implementation on a Vaillant heat pump. 

The only small thing that I have noticed was the error returned if the user is providing an invalid measurement type.

usecases/ma/mdt/public.go
andig: Updated, including neighboring use cases, too. evcc has already been running MDT for a few weeks now (Vaillant and Wolf).

andig:
PS.: don't forget to squash when merging ;)

Ka0o0 (approved):
Thanks, looks good

PR #228, opened 2026-06-25
Register OHPCF event handler in constructor

andig:
`NewOHPCF` is the only use-case constructor in this repo that does not subscribe the concrete use case to the device event bus. `NewUseCaseBase` only subscribes the embedded `*UseCaseBase`, whose `HandleEvent` handles use-case-data updates (firing `UseCaseSupportUpdate`). `OHPCF.HandleEvent` — which on entity-add calls `connected()` to `Subscribe()`/`Bind()` the remote `SmartEnergyManagementPs` feature, and on data change runs `loadSmartEnergyManagementPsDataType` to emit the `DataUpdate*` events — is therefore never registered and never runs.

Add the same `_ = localEntity.Device().Events().Subscribe(uc)` call to `NewOHPCF`.

Also added requesting initial data same als LPC does.

### DIFF
--- a/usecases/cem/ohpcf/events.go
+++ b/usecases/cem/ohpcf/events.go
@@ -43,6 +43,11 @@ func (o *OHPCF) connected(entity spineapi.EntityRemoteInterface) {
 				logging.Log().Debug(err)
 			}
 		}
+
+		// read the current data as a subscription only delivers future updates
+		if _, err := semp.RequestData(); err != nil {
+			logging.Log().Debug(err)
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


andig:
@sthelen-enqs this is an actual bug with https://github.com/enbility/eebus-go/pull/223

usecases/cem/ohpcf/events.go
andig, line 121:
```suggestion
```

sthelen-enqs:
Thanks for the fixes and for testing them with a real device. Sorry for the delay in reviewing and merging this.

I wasn't expecting the merge of #223 to close this PR before I could merge it. I grabbed the patches for this from GitHub and will merge this in #230 (your attribution should still be there)

Issue #173, opened 2025-05-06
EG LPC should be able to notify nominal max value changes


DerAndereAndi:
The controllable system can set nominal max values. The EG LPC implementation has a getter method to return the current value, but it does not provide notifications and the subscription to receive any updates for them.

sthelen-enqs:
Subscriptions are added by #189, the only missing step is to create notifications for updates to the nominal max values.

msbakhshi:
Changes added https://github.com/enbility/eebus-go/pull/205

PR #229, opened 2026-06-26
EG LPC: subscribe and read electrical connection characteristics

### LINKED ISSUE #225: uccslpc.SetConsumptionNominalMax is not seen by the energy guard
I used the hems example from the examples folder. It calls SetConsumptionNominalMax before the service gets started. The problem is that ConsuptionNominalMax ist not seen be the energy guard at the other side (Error getting ConsumptionNominalMax: data not available)
If I call SetConsumptionNominalMax at the event of UseCaseSupportUpdate, the other end sees it sometimes. If I call it after a Heartbeat, it is seen be the energy guard.

andig:
## Problem

The EG LPC use case has a `ConsumptionNominalMax` getter, but `connected()` only set up subscriptions/reads for LoadControl, DeviceConfiguration and DeviceDiagnosis — never for ElectricalConnection. As a result the value is only ever populated by an unsolicited partial notify from the controllable system, so:

- a value `Set` before a subscriber exists is never delivered (#225), and
- there is no subscription/notification for changes (#173).

The getter therefore returns `ErrDataNotAvailable` until the CS happens to re-send the characteristic to an already-connected subscriber.

## Fix

- On connect, subscribe to the remote ElectricalConnection feature and read its characteristics, so the current nominal max is fetched immediately.
- Dispatch `ElectricalConnectionCharacteristicListData` updates and emit a new `DataUpdateConsumptionNominalMax` event when the nominal max becomes available.

Mirrors the existing DeviceConfiguration handling. Test added covering both the dispatch case and the handler.

Fixes #225, #173

### DIFF
--- a/usecases/eg/lpc/events.go
+++ b/usecases/eg/lpc/events.go
@@ -41,6 +41,9 @@ func (e *LPC) HandleEvent(payload spineapi.EventPayload) {
 
 	case *model.DeviceConfigurationKeyValueListDataType:
 		e.configurationDataUpdate(payload)
+
+	case *model.ElectricalConnectionCharacteristicListDataType:
+		e.electricalConnectionCharacteristicDataUpdate(payload)
 	}
 }
 
@@ -91,6 +94,19 @@ func (e *LPC) connected(entity spineapi.EntityRemoteInterface) {
 		}
 	}
 
+	if electricalConnection, err := client.NewElectricalConnection(e.LocalEntity, entity); err == nil {
+		if !electricalConnection.HasSubscription() {
+			if _, err := electricalConnection.Subscribe(); err != nil {
+				logging.Log().Debug(err)
+			}
+		}
+
+		// get the current characteristics, e.g. the nominal max consumption
+		if _, err := electricalConnection.RequestCharacteristics(nil, nil); err != nil {
+			logging.Log().Debug(err)
+		}
+	}
+
 	if deviceDiagnosis, err := client.NewDeviceDiagnosis(e.LocalEntity, entity); err == nil {
 		if !deviceDiagnosis.HasSubscription() {
 			if _, err := deviceDiagnosis.Subscribe(); err != nil {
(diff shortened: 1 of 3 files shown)


FernetMenta:
I tested in my environment and can confirm that it fixes #225

sthelen-enqs:
PR https://github.com/enbility/eebus-go/pull/189 which we just merged added the subscription and read to the LPC and LPP use cases for scenario 4 which should cover the reads of the characteristics. Your pull request also adds a notification for LPC which we've agreed would be nice to have as well. Could you rebase this PR on dev and add the  electricalConnectionCharacteristicDataUpdate function to the LPP use case as well?
If not, we can close this PR and we'll take a look at adding the notification on our side.

sthelen-enqs:
Actually, this should already be part of https://github.com/enbility/eebus-go/pull/205 which was just merged. Could you test with the actual dev and see if changes are still required?

FernetMenta:
I tested with branch dev and it does not fix #225 

```
2026-06-29 14:05:48 TRACE 8c61626a66112125d1c21f71f9b213f095d649f9 Send: 8c61626a66112125d1c21f71f9b213f095d649f9 {"messageProtocolHandshake":[{"handshakeType":"announceMax"},{"version":[{"major":1},{"minor":0}]},{"formats":[{"format":["JSON-UTF8"]}]}]}
Error getting ConsumptionNominalMax: data not available
2026-06-29 14:05:48 TRACE 8c61626a66112125d1c21f71f9b213f095d649f9 Recv: 8c61626a66112125d1c21f71f9b213f095d649f9 {"data":[{"header":[{"protocolId":"ee1.0"}]},{"payload":{"datagram":[{"header":[{"specificationVersion":"1.3.0"},{"addressSource":[{"device":"d:_n:Demo_HEMS-123456789"},{"entity":[1]},{"feature":5}]},{"addressDestination":[{"device":"d:_n:DemoVendor_Device-Tester-Demo-HEMS-123"},{"entity":[1]},{"feature":4}]},{"msgCounter":18},{"msgCounterReference":18},{"cmdClassifier":"reply"}]},{"payload":[{"cmd":[[{"electricalConnectionCharacteristicListData":[{"electricalConnectionCharacteristicData":[[{"electricalConnectionId":0},{"parameterId":0},{"characteristicId":0},{"characteristicContext":"entity"},{"characteristicType":"contractualConsumptionNominalMax"},{"value":[{"number":32000},{"scale":0}]},{"unit":"W"}],[{"electricalConnectionId":0},{"parameterId":0},{"characteristicId":1},{"characteristicContext":"entity"},{"characteristicType":"contractualProductionNominalMax"},{"value":[{"number":10000},{"scale":0}]},{"unit":"W"}]]}]}]]}]}]}}]}
EgLPC Event:  eg-lpc-DataUpdatePowerConsumptionNominalMax
Error getting ConsumptionNominalMax: no compatible entity
```

andig:
Can you share the full log?

FernetMenta:
Sorry for the noise, dev does fix #225. I missed to trigger on new event "DataUpdatePowerConsumptionNominalMax", hence my results were not reliable before.

sthelen-enqs:
Thanks, I'll close this then. Feel free to open an issue if you notice something that isn't working correctly

PR #231, opened 2026-07-07
Add msgCounter to OHPCF result callbacks

andig:
Follow-up to #210 as requested in https://github.com/enbility/eebus-go/pull/210#issuecomment-4901842186.

#210 added `msgCounter` to the result callback for `WriteLoadControlLimits` (OPEV/OSCEV) and `WriteConsumptionLimit`/`WriteProductionLimit` (EG LPC/LPP), but didn't touch OHPCF's `SchedulePowerConsumptionProcess`/`AbortPowerConsumptionProcess`/`PausePowerConsumptionProcess`/`ResumePowerConsumptionProcess`, leaving the result-callback shape inconsistent across usecases (OHPCF still only gets `func(result model.ResultDataType)`).

This applies the same change to OHPCF's four write methods and their shared `writeSmartEnergyManagementData` helper, mirroring #210's pattern exactly.

### DIFF
--- a/usecases/api/cem_ohpcf.go
+++ b/usecases/api/cem_ohpcf.go
@@ -72,14 +72,14 @@ type CemOHPCFInterface interface {
 	//
 	// parameters:
 	//   - startIn: Delay from now until the power consumption starts (0 = start immediately)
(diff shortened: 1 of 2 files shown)


andig:
@sthelen-enqs also had to gofmt the file which #223 omitted.

sthelen-enqs (approved):
Thanks for this, the two branches for this and OHPCF were run separated for too long so we forgot to update this for OHPCF

PR #236, opened 2026-07-08
Fix validActorTypes for GCP MGCP use case.

sthelen-enqs:
validActorTypes contains the valid remote actors, not which actors are valid locally.


andig:
To make this more robust in future, consistenly rename the variable to `validRemoteActorTypes` everywhere?

sthelen-enqs:
It's probably a good idea to change that to lower the chances of developer confusion. We'll look into updating this once the other usecase PRs are merged

PR #239, opened 2026-07-17
Add Hvac and Setpoint client features

andig:
Adds `features/client` and `features/internal` helpers for the SPINE `HVAC` and `Setpoint` features, including tests.

This is groundwork for the remaining HVAC use cases (system function monitoring/configuration and temperature configuration), which will be submitted as separate PRs per use case, stacked on this one:

- Hvac: read system function descriptions/data, operation mode descriptions, operation mode and setpoint relations; write `hvacSystemFunctionListData`
- Setpoint: read setpoint descriptions/data/constraints; write `setpointListData`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #244, opened 2026-07-17
Add Configuration of Room Cooling System Function (CRCSF) use case

andig:
Adds the **Configuration of Room Cooling System Function (CRCSF)** use case (EEBus UC TS V1.0.0) as a **Configuration Appliance** client actor in `usecases/ca/crcsf`.

Scenario 1 (mandatory): set the room cooling operation mode on an `HVACRoom` entity.

Structural mirror of the CRHSF use case (#241) for the **cooling** system function.

Public API: `OperationModes`, `CurrentOperationMode`, `WriteOperationMode` (gated by `isOperationModeIdChangeable`).

Stacked on #239 (Hvac and Setpoint client features) — only the last commit is specific to this PR.


andig:
ping @volschin refs https://github.com/enbility/eebus-go/pull/247#issuecomment-5024156516

PR #249, opened 2026-07-20
Add Configuration of DHW Temperature (CDT) use case

volschin:
## Summary

Adds the Configuration Appliance client for **Configuration of DHW Temperature (CDT)** and supersedes the older draft #132 with the current use-case APIs and event lifecycle.

This contribution is stacked on #239 because CDT uses its shared HVAC API types and the Setpoint/HVAC client helpers. The CDT-specific change is commit `50de294`.

## Behavior and safety

- resolves setpoints only through the DHW system-function relations
- accepts any advertised operation mode, including `auto` when it maps unambiguously to exactly one DHW setpoint
- checks the remote `SetpointListData` write operation and the optional changeability flag
- validates complete min/max/step constraints before writing
- preserves unrelated entries by sending a full cached setpoint-list copy
- exposes the message counter and the remote result callback
- refreshes setpoint data after a successful device result
- fails closed for ambiguous DHW functions, relations, incomplete constraints, and duplicate target entries

The full-list write and omitted-changeability behavior reflect the interoperability behavior observed with a Vaillant VR940 gateway and retain the validation semantics already exercised in the eebus-bridge integration.

## Verification

- `go vet ./...`
- `go test -race ./...`


volschin:
Ping @andig

andig:
Yes?

volschin:
The missing part to complete DHW.

andig:
Ah. Nothing evcc uses ;). I'd go for `cdsf.StartOneTimeDhw`.

volschin:
The Start.OneTimeDhw only heats the water to the configured standard temp. This gives the Option to higher the temp for this action and set it back afterwards.

PR #253, opened 2026-07-23
fix: ignore useCaseAvailable flag on client (EG) side

### LINKED ISSUE #252: Bug: Client aborts use case setup when useCaseAvailable=false (TC_SPINE_RTC_003)
**Test case:** TC_SPINE_RTC_003 — Ignore useCaseAvailable flag  
**Reference:** [SPINE-TS-UCD-01]  
**Severity:** Non-conformance (interoperability failure)

## Description

TC_SPINE_RTC_003 requires that the DUT (acting as Energy Guard / client) **completely ignores** the `useCaseAvailable` flag received from a remote server during Use Case discovery. Even if the flag is explicitly set to `false`, the DUT MUST silently ignore it and proceed with the control flow by successfully initiating the required node management operations.

eebus-go violates this requirement. When processing incoming `nodeManagementUseCaseData`, the generic use case event handler checks `useCaseAvailable` and **skips** the use case support entry if the flag is `false`. This causes the client to abort use case setup instead of proceeding.

## Non-conformant code

**`usecases/usecase/events.go:61–66`**

```go
if support.UseCaseName == nil ||
    *support.UseCaseName != u.UseCaseName ||
    support.ScenarioSupport == nil &&
        len(support.ScenarioSupport) == 0 ||
    (support.UseCaseAvailable != nil && !*support.UseCaseAvailable) {  // <-- violates TC_SPINE_RTC_003
    continue
}
```

The condition on line 65 causes the handler to `continue` (skip this use case support entry) whenever the remote partner sends `useCaseAvailable=false`. As a result, the client never calls the downstream entity-discovery logic, never registers remote entities for the use case, and never initiates node management operations — failing the test case.

The existing comment at lines 58–60 acknowledges some ambiguity ("UseCaseAvailable should only be checked for client functionality...") but the resolution chosen (check it always) is incorrect per [SPINE-TS-UCD-01]: clients must not gate on this flag at all.

## Required fix

Remove the `useCaseAvailable` check from the guard condition at `usecases/usecase/events.go:65`:

```go
if support.UseCaseName == nil ||
    *support.UseCaseName != u.UseCaseName ||
    support.ScenarioSupport == nil &&
        len(support.ScenarioSupport) == 0 {
    continue
}
```

The `useCaseAvailable` flag must be entirely disregarded on the client (EG) side. If there is a separate need to handle it for server-side logic (as the comment hints at for Porsche PMCC), that must be handled in a different, server-specific code path — not in the shared client event handler.

## Affected use cases

All use cases using the shared `UseCaseBase` event handler are affected when acting in the EG (client) role, including:

- LPC (Load Power Control) — `usecases/cem/lpc/`
- LPP (Limitation of Power Production) — `usecases/cem/lpp/`
- Any other use case routed through `usecases/usecase/events.go:useCaseDataUpdate()`

## Addendum: ScenarioSupport condition is also incorrect

The `ScenarioSupport` guard at `usecases/usecase/events.go:63–64` has a logic bug independent of the `useCaseAvailable` issue:

```go
support.ScenarioSupport == nil &&
    len(support.ScenarioSupport) == 0 ||
```

The correct semantics are:
- `ScenarioSupport == nil` — field omitted entirely → skip the entry (no information)
- `ScenarioSupport == []` (empty non-nil) — intentional "no scenarios supported" → proceed and clear any existing registration

The current condition `ScenarioSupport == nil && len(ScenarioSupport) == 0` is **functionally correct** but misleading: in Go `len(nil) == 0` always, so the `&& len(...) == 0` part is unconditionally true whenever the nil check is true — making it a redundant dead clause. The condition is equivalent to the simpler and clearer:

```go
support.ScenarioSupport == nil ||
```

The `Required fix` section above should use the corrected form (nil-only check, no redundant len):

```go
if support.UseCaseName == nil ||
    *support.UseCaseName != u.UseCaseName ||
    support.ScenarioSupport == nil {
    continue
}
```

This is a code clarity issue, not a functional bug.

## Test scenario

Preconditions: `PRE_TestTool_ReplyUcdAvailableFalse_Configured`, `PRE_SPINE_NewConnectionEstablished`

Step 1: Test tool (CS) responds to Use Case discovery with `useCaseAvailable=false` for the relevant use case.  
Expected: DUT (EG) ignores the flag and successfully completes `PRE_SPINE_Scenario_LpcLpp_S1_S3` within 120s.  
Actual: DUT silently drops the use case, never initiates node management operations → scenario fails.

Test case that should reproduce the issue: [bug_TC_SPINE_RTC_003.patch](https://github.com/user-attachments/files/30266398/bug_TC_SPINE_RTC_003.patch)

TestSpec Analysis done by me, ticket description done by Claude. Detailed reasoning and required fix may require further manual analysis.

andig:
Fixes #252 (TC_SPINE_RTC_003).

## Problem

Per SPINE-TS-UCD-01, a client acting as Energy Guard must **completely ignore** the `useCaseAvailable` flag received during Use Case discovery. The shared `UseCaseBase` event handler (`usecases/usecase/events.go`) instead gated on it and `continue`d past the use case support entry whenever the remote server sent `useCaseAvailable=false` — so the client aborted setup and never initiated node management operations. Affects all use cases routed through `useCaseDataUpdate()` (LPC, LPP, ...) in the EG role.

## Fix

- Remove the `useCaseAvailable` check from the guard condition.
- Simplify the redundant `ScenarioSupport == nil && len(...) == 0` to `ScenarioSupport == nil` (`len(nil) == 0` always).
- Replace the stale comment referencing the old behavior.

## Test

Updated `events_test.go`: the EV support entry with `useCaseAvailable=false` now asserts the use case registers instead of being dropped. All `usecases/...` suites pass, including `eg/lpc` and `eg/lpp`.

### DIFF
--- a/usecases/usecase/events.go
+++ b/usecases/usecase/events.go
@@ -55,14 +55,12 @@ func (u *UseCaseBase) useCaseDataUpdate(
 		}
 
 		for _, support := range uc.UseCaseSupport {
-			// UseCaseAvailable should only be checkd for client functionality
-			// of a use acse. But as there are devices (e.g. Porsche PMCC) that
-			// also use it for the server side, we need to check it always.
+			// Per SPINE-TS-UCD-01 (TC_SPINE_RTC_003), the client (EG) side must
+			// ignore the UseCaseAvailable flag entirely and proceed with setup,
+			// even when the remote server reports UseCaseAvailable=false.
 			if support.UseCaseName == nil ||
 				*support.UseCaseName != u.UseCaseName ||
-				support.ScenarioSupport == nil &&
-					len(support.ScenarioSupport) == 0 ||
-				(support.UseCaseAvailable != nil && !*support.UseCaseAvailable) {
+				support.ScenarioSupport == nil {
 				continue
 			}
 
--- a/usecases/usecase/events_test.go
+++ b/usecases/usecase/events_test.go
@@ -66,6 +66,8 @@ func (s *UseCaseSuite) Test_useCaseDataUpdate() {
 	result = s.uc.IsScenarioAvailableAtEntity(s.monitoredEntity, 1)
 	assert.False(s.T(), result)
 
+	// TC_SPINE_RTC_003: even with useCaseAvailable=false the client (EG) must
+	// ignore the flag and register the use case support.
 	data = &model.NodeManagementUseCaseDataType{}
 	data.AddUseCaseSupport(
 		*feature.Address(),
@@ -80,7 +82,7 @@ func (s *UseCaseSuite) Test_useCaseDataUpdate() {
 	s.uc.useCaseDataUpdate(payload)
 
 	result = s.uc.IsScenarioAvailableAtEntity(s.monitoredEntity, 1)
-	assert.False(s.T(), result)
+	assert.True(s.T(), result)
 
 	data = &model.NodeManagementUseCaseDataType{}
 	data.AddUseCaseSupport(


usecases/usecase/events.go
```diff
@@ -55,14 +55,12 @@ func (u *UseCaseBase) useCaseDataUpdate(
...
-			// also use it for the server side, we need to check it always.
+			// Per SPINE-TS-UCD-01 (TC_SPINE_RTC_003), the client (EG) side must
+			// ignore the UseCaseAvailable flag entirely and proceed with setup,
+			// even when the remote server reports UseCaseAvailable=false.
 			if support.UseCaseName == nil ||
 				*support.UseCaseName != u.UseCaseName ||
-				support.ScenarioSupport == nil &&
-					len(support.ScenarioSupport) == 0 ||
-				(support.UseCaseAvailable != nil && !*support.UseCaseAvailable) {
+				support.ScenarioSupport == nil {
```
sthelen-enqs, line 63: I don't know if we should check ScenarioSupport == nil or len(ScenarioSupport) == 0 here.
I think I'll merge it as is because that way at least we're not changing our behavior here.

sthelen-enqs (approved):

usecases/usecase/events.go
andig: In Go `len(nil) == 0`. This library generally makes that confusing and unnecessarily complex. Since cosmetic PRs are not desired I've never opened one but the pattern should be globally avoided.

sthelen-enqs: Yeah, I was more generally wondering what we should do if the list exists but is empty.

